PR #76, opened 2026-06-18
Automatically set thresholds based on data

### LINKED ISSUE #68: Adaptive confidence and min size thresholds
For 25% scale Library of Congress images, I typically use min-short-side=20, min-long-side=40, min-confidence=0.15. But this isn't alway the best choice. Sometimes lowering these gains critical street labels.

Examples:

- Queens 1950 vol 2
  - p41 & p62: lowering the confidence threshold to 0.10 or 0.07 would unlock numbered streets in these images.
  - p48: lowering the size threshold would unlock several critical, high-confidence street detections.
- Queens 1950 vol 5 is a complete disaster, the writing is all too small for the threshold.

This could work well with #50. If we know roughly where the map should be, and how it should be oriented, then it's safer to try lowering the threshold.

danvk:
Fixes #68.

Different scanned Sanborn volumes have very different effective OCR text sizes and confidence levels — a single hard-coded `--min-confidence`/`--min-long-side`/`--min-short-side` default is too strict for some volumes (missing detections, no fit) and fine for others. This replaces the earlier per-page "adaptive escalation" idea (abandoned — it almost never changed the outcome for an individual page) with **volume-wide auto-calibration**: before processing a volume, `mapsnap fit`/`mapsnap georef` now picks thresholds based on that volume's own detection statistics, unless the user passes one of the three flags explicitly (manual override always wins and skips calibration entirely).

## How thresholds are picked

Each of `min_confidence`, `min_long_side`, `min_short_side` is calibrated **independently** — they are not proportional to each other, and don't move by the same amount or in the same direction. What they do share is a single relaxation level: `calibrate_thresholds` evaluates today's CLI default plus four percentile rungs (`ADMIT_FRACTION_LEVELS = [40, 55, 70, 85]`), and at a given rung, each of the three thresholds is independently set to the value that admits the top N% of *that metric's own* distribution across the volume's cached detections (`detection_percentiles` collects per-detection confidence/long_side/short_side, `thresholds_at_admit_fraction` takes the N-th percentile of each list separately). So at "rung 70", confidence might barely move while short_side drops a lot, or vice versa — whatever each metric's own distribution calls for. This is also why the absolute thresholds self-scale across scans of different resolution: a volume with small detected text gets small absolute cutoffs, one with large text gets large ones, automatically.

For each candidate rung (5 total, including the default), calibration measures the fraction of pages with at least `--min-gcps-for-calibration` (default 2) candidate intersection GCPs — the real signal for whether a page can fit, since a page can have plenty of qualifying detections and still fail if none of them form an intersecting pair. It picks whichever rung maximizes that fraction; ties favor the strictest (least-relaxed) rung, so an already-fine volume is never gratuitously loosened. This is a plain best-of-5 search rather than "walk outward until good enough," because the fraction isn't guaranteed to be monotonic in relaxation level — one volume's fraction measurably *dipped* at one percentile rung before recovering at a looser one.

## Calibration cost vs. a full fit

Calibration is cheap relative to the actual fit. `candidate_gcps_for_page` (used both by calibration and by `process_image`'s real run, so they can never drift apart) mirrors `process_image`'s preprocessing exactly — ignore/hint filtering, avenue-letter promotion, dedup, confidence/size/aspect filtering, canonical street-name resolution, square-feature direction correction — and calls the existing intersection-lookup, but stops there: no RANSAC (no O(n²) search over candidate GCP pairs), no affine fitting/inlier scoring, no image decode, no file writes.

Measured on a 87-page volume: the full 5-rung calibration sweep (i.e. 5 full passes over every page, each just the cheap preprocessing step above) took ~84s. The subsequent real run — RANSAC + writing output for every page, once — took ~370s. So calibration is roughly 5 cheap passes adding up to less than a quarter of total runtime, in exchange for not running the expensive RANSAC search 5x.

## Other changes in this PR

While verifying this end-to-end, the calibrated thresholds for some volumes legitimately admit far more raw detections per page, which exposed a pre-existing O(n²) blowup in `ransac_hybrid`'s candidate-pair search (one volume's run climbed past 70+ CPU-minutes). Fixed in two parts:
- `ransac_hybrid` now takes a `max_gcps` parameter (default 40) and truncates its (already closest-pair-first sorted) candidate list before the O(n²) pair search.
- `project_to_polyline` — called by RANSAC's inlier scoring for every (candidate affine, feature) pair, and the actual dominant cost per cProfile — was rewritten from an unvectorized per-segment Python loop into a handful of batched numpy operations. Verified byte-for-byte equivalent to the original via 2000 randomized samples plus direct in-process comparison.

Together these brought one volume's full auto-calibrated run from ~75 minutes down to ~6 minutes, with no change to fit outcomes (verified against the original implementation in isolation; small page-count differences seen across repeated runs are a pre-existing, unrelated nondeterminism from Python's hash-seed randomization affecting label-dedup ordering — reproduced on the pre-existing code too, not introduced here).

## Verification

Ran end-to-end (auto vs. `--disable-auto-threshold` baseline) on all four reference volumes:

| volume | baseline | auto |
|---|---|---|
| champaign_ill_1915 | 27/33 | 27/33 (stays at default, as expected) |
| chicago_il_1950_vol_1 | 65/111 | 104/111 |
| queens_1950/vol2 | 67/88 | 72–77/88 (small run-to-run variance from the pre-existing nondeterminism above) |
| queens_1950/vol5 | 1/68 | 49/68 |

`uv run ruff format/check`, `uv run pyright`, and `uv run pytest` (277 tests) all pass.

### DIFF
--- a/mapsnap/georef_from_labels.py
+++ b/mapsnap/georef_from_labels.py
@@ -62,6 +62,15 @@ class IntersectionGCP:
 
 
 @dataclass
+class Thresholds:
+    """Detection-acceptance thresholds passed to process_image / candidate_gcps_for_page."""
+
+    min_confidence: float
+    min_long_side: float
+    min_short_side: float
+
+
+@dataclass
 class ProcessResult:
     """Outcome of processing a single map image."""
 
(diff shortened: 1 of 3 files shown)


danvk:
Closed in favor of #77, which is simpler.

PR #97, opened 2026-07-01
Place sole-intersection jog pages via 1-GCP fit instead of failing

danvk:
GCPs that share one image pixel are the same intersection with alternative world locations — a jogging street crossing its cross-street twice, or a divided road's two carriageways. They can't pair with each other to fix an affine, so a page whose GCPs all collapse onto one pixel has just **one** control point.

Rather than fail such a page, defer it to the median-scale 1-GCP fit, which now tries **each** world candidate at the shared pixel and keeps the one with the most inlier labels (then lowest total residual). Adds `_distinct_pixel_gcps` and `_inlier_residuals`.

Verified: full test suite green.

## Experiment — 6 volumes vs `main` (f62de9b9)

Each row: `mapsnap fit` on this branch vs main; mean RMSE (ft) against OIM truth, georeferenced-page count, and notable per-page changes (≥20 ft or coverage).

| Volume | mean RMSE ft (Δ) | georef pages | notable |
|---|---|---|---|
| brooklyn_ny_1939_vol_1 | 20.8 → 20.8 (+0.0) | 54 → 54 | no change |
| champaign_ill_1915 | 13.0 → 13.0 (+0.0) | 18 → 18 | no change |
| chicago_il_1950_vol_1 | 114.1 → 114.1 (+0.0) | 98 → 98 | no change |
| detroit_mich_1929_vol_11 | 32.8 → 33.1 (+0.3) | 83 → 84 | placed p32 |
| new_orleans_la_1896_vol_2 | 53.1 → 53.1 (+0.0) | 80 → 80 | no change |
| new_orleans_la_1951_vol_5 | 17.0 → 17.3 (+0.3) | 102 → 103 | placed p432 |

**Takeaway:** places 2 previously-failing pages (Detroit p32, New Orleans-1951 p432) at a negligible +0.3 ft mean cost on those volumes; every other volume unchanged. (Chicago's 114 ft is a property of the `main` baseline — identical on both sides.)

### DIFF
--- a/mapsnap/georef_from_labels.py
+++ b/mapsnap/georef_from_labels.py
@@ -521,6 +521,61 @@ def find_intersection_gcps(
     return sorted(gcps, key=lambda g: g.pixel_dist)
 
 
+# GCPs whose pixel crossings fall within this many pixels are treated as one image
+# intersection with several world candidates (a jogging street's two crossings, or a
+# divided road's two carriageways), not as independent control points.
+SAME_PIXEL_TOL_PX = 5.0
+
+
+def _distinct_pixel_gcps(
+    gcps: list[IntersectionGCP],
+) -> list[list[IntersectionGCP]]:
+    """Group GCPs by near-coincident pixel crossing.
+
+    Each group is a set of GCPs sharing one image point, whose geo locations are therefore
+    alternatives (e.g. the two crossings of a jogging street). The number of groups is the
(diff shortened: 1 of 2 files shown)


danvk:
This is basically a bug fix. The two new fits are both good. It's more code than I would have expected, but 🤷

Issue #51, opened 2026-06-10
Require support for ambiguous words like CANAL


danvk:
There are lots of words that appear in street names that can appear in other contexts in Sanborn maps:

- RIVER
- BRIDGE
- CANAL
- WATER
- PARK
- CONGRESS (as in "Library of Congress")
- SANBORN
- (name of city)

These should be detected specifically and thrown out unless they're supported by a "hint," e.g. "CANAL STREET" is fine but a plain "CANAL" is not.

Context: a bad "CANAL" detection in Chicago p106w produces a poor fit.

danvk:
Using the keymap to filter to just the streets in the page's area proved to be a much better fix. #102

PR #100, opened 2026-07-02
Filter ambiguous bare words (CANAL, CHICAGO, …) without a street-type hint

danvk:
Bare geographic/label words that also name a street — `RIVER`, `BRIDGE`, `CANAL`, `WATER`, `PARK`, `CONGRESS`, `SANBORN`, plus the volume's own city name (e.g. `CHICAGO` on a Chicago sheet, from "Chicago River"/"City of Chicago") — are frequent false street matches: they label a feature, not a street. A bare one is now trusted only when a `STREET`/`AVENUE` type-word hint sits next to it in the same column (`CANAL` alone → filtered; `CANAL` + `ST` → kept). Implements issue #51.

`AMBIGUOUS_STREET_WORDS` + `is_ambiguous_word` / `city_ambiguous_words` live in `streets.py`; `has_type_hint_support` and the detection-filter hook are in `georef_from_labels.py`.

Verified: full test suite green.

## Experiment — 6 volumes vs `main` (f62de9b9)

Each row: `mapsnap fit` on this branch vs main; mean RMSE (ft) against OIM truth, georeferenced-page count, and notable per-page changes (≥20 ft or coverage).

| Volume | mean RMSE ft (Δ) | georef pages | notable |
|---|---|---|---|
| brooklyn_ny_1939_vol_1 | 20.8 → 20.1 (-0.7) | 54 → 53 | dropped p9 |
| champaign_ill_1915 | 13.0 → 13.6 (+0.6) | 18 → 18 | no change |
| chicago_il_1950_vol_1 | 114.1 → 14.0 (-100.1) | 98 → 93 | dropped p48N, p53N, p87W, p88W, p94W; p92W 83→13ft; p89W 96→42ft; p6N 7→51ft |
| detroit_mich_1929_vol_11 | 32.8 → 43.5 (+10.7) | 83 → 84 | placed p93 |
| new_orleans_la_1896_vol_2 | 53.1 → 52.0 (-1.1) | 80 → 79 | dropped p105; p183 71→20ft; p103 40→13ft |
| new_orleans_la_1951_vol_5 | 17.0 → 17.0 (+0.0) | 102 → 102 | no change |

**Takeaway.** The headline is **Chicago −100 ft** (114 → 14): `main` was placing two pages catastrophically wrong — p53N seeded off `CANAL`/`WATER` (Lake Michigan labels → Canal St) and p48N off `CHICAGO` (the city name → E. Chicago Ave); the filter discards those false GCPs (5 junk pages dropped, several others improved). NOLA-1896 also nets −1.1 ft.

Two costs, both expected:
- **Brooklyn p9 dropped** — a *legitimate* CONGRESS Street with no adjacent `STREET` hint. On `main` p9 fits at 59.2 ft seeded by `CONGRESS × COLUMBIA`; the filter removes bare `CONGRESS`, so p9 falls to misscale. This is the precision/recall cost of the blacklist: it can't tell a real hint-less Congress St from a spurious one. (The volume mean nominally drops 0.7 ft only because p9 was above-average error — it's a coverage loss, not a win.)
- **Detroit +10.7 ft** — a *side effect*, not a direct filtering error: removing ambiguous detections shifts the volume's median scale, and p93 (a misscale outlier on `main`) slips back under the scale-outlier threshold as a 913 ft garbage placement. Same median-scale coupling seen elsewhere; orthogonal to this change's intent.

Net strongly positive (Chicago dominates), but the Brooklyn p9 CONGRESS false-negative is a known limitation worth a follow-up (e.g. also accept an ambiguous word that is an unambiguous, unique street match in the volume's centerlines).

### DIFF
--- a/mapsnap/georef_from_labels.py
+++ b/mapsnap/georef_from_labels.py
@@ -35,7 +35,9 @@ from mapsnap.streets import (
     build_block_index,
     canonical_street_matches,
     deduplicate_detections,
+    city_ambiguous_words,
     hint_type_word,
+    is_ambiguous_word,
     is_bare_letter,
     is_number_only,
     normalize_street,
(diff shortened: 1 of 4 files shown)


danvk:
Closing in favor of #102, which is more targeted

PR #103, opened 2026-07-04
Architectural swing

danvk:
Prototype of **predict-then-verify** georeferencing (`mapsnap/predict_fit.py`, standalone CLI — not wired into the pipeline): instead of match-then-fit (label names → intersection GCPs → RANSAC), start from the prior transform the key map implies for every placed page and let names *verify* the geometry. The goal was to prove the concept on real volumes, get concrete numbers, and surface unforeseen limitations. All three happened.

## Architecture

A page's prior similarity transform comes entirely from key-map artifacts built on #102:

- **Translation** — the page number's detection center(s) on the georeferenced key map (one per split panel; nearest-center semantics, never the mean).
- **Scale** — the volume's median fitted scale × the page's region multiplier (a half-scale sheet has a ~2×-area key-map region), i.e. the same prior that drives #102's deferred-fit snap.
- **Rotation** — *voted, not assumed* (truth data shows volume rotation is bimodal in Chicago and arbitrary in Brooklyn/New Orleans): every (label, same-name candidate segment) pair votes θ = bearing + label-direction (mod π); correct pairs pile onto a mode, wrong candidates scatter. Top two modes × both half-turns = up to 4 initializations.

Fitting then iterates association ↔ weighted least squares over a shrinking tolerance schedule (1500 → 250 ft):

- **Association**: project each label, take the nearest same-name segment within tolerance whose bearing agrees with the label's predicted direction (±30°).
- **Solve**: each match contributes a *point-to-line* row (label lies on its street — no intersections needed, so a page with two street labels and no usable crossing is already over-determined) **and a direction row** (the label's mapped direction stays parallel to its street — linear in the transform). Weak rows anchor the page center to the key-map location; scale is clamped to the prior band.
- **Verify, or refuse**: a fit is emitted only if ≥3 label inliers (<100 ft) across ≥2 street names whose normals span ≥30°, inlier RMS ≤ 60 ft, label spread ≥ 25% of the image diagonal, fitted center within 1.5 radii of the key-map location, and fitted scale within **[0.85, 1.2]** of the prior.

## Methodology

- Volumes: **Detroit 1929 v11** (development) and **Chicago 1950 v1** (held-out validation — dense grid, the hardest association-ambiguity test). Verification thresholds were tuned on Detroit and *frozen* before Chicago ran.
- Truth eval: per-page RMSE over a 7×7 pixel grid against the OIM truth annotations, computed identically for the prototype and the current pipeline's in-place fits. Split panels are excluded from truth scoring (their truth GCPs live in full-sheet coordinates; the official compare resolves this by polygon IOU, this evaluator does not).
- Selection-rule experiment: candidate merge rules tuned on Detroit, frozen, validated on Chicago (tables below).
- Runtime: ~40 s per ~100-page volume, single-threaded.

## Results

On pages where **both** fitters place (truth-scored, unsplit):

| | Detroit (n=50) | Chicago (n=62) |
|---|---|---|
| prototype median / mean / max | 12.1 / **16.0** / **74** ft | 11.6 / 18.6 / 117 ft |
| pipeline median / mean / max | 11.6 / 19.5 / 166 ft | **10.5 / 14.3 / 83** ft |
| prototype better / worse (>5 ft) | 10 / 10 | 7 / 12 |

- **Detroit**: ties the mature intersection fitter at the median and beats its tail — p44 166→21 ft, p51 38→10, p48 50→24, p71 84→63, p46 32→15, p43 85→74. Verified 52/101 attempted.
- **Chicago**: verified 64/100 with **no catastrophes** (max 117 ft on the ambiguity-stress volume — the refuse-first verification held), but slightly worse than the pipeline overall: exact intersections beat point-to-line residuals where the grid is dense. Losses are 20–80 ft drift, not branch-jumps.
- Gap fills (pages the pipeline can't place at all): Detroit +2 split panels, Chicago +2 (one split panel; p50N at 196 ft).

## Combined-selection rule

Rules evaluated over truth pages (pick prototype vs pipeline per page, no truth access in the rule):

| rule | Detroit median/mean/max | Chicago median/mean/max (frozen) |
|---|---|---|
| pipeline only | 14.0 / 25.5 / 246 | **9.7 / 15.5 / 96** |
| S1: + prototype where pipeline empty | 14.0 / 25.5 / 246 | 9.8 / 17.6 / 196 (+1 page) |
| S2: also replace deferred 1-GCP fits | 14.0 / 25.6 / 246 | 9.8 / 17.6 / 196 |
| S3: also replace when inliers ≥4, rms ≤30 | **12.6 / 22.3 / 246** | 10.4 / 17.9 / 196 |
| oracle (min per page) | 12.1 / 21.5 / 246 | 8.9 / 16.2 / 196 |

Verdict: **S3-style replacement is not licensed cross-volume** — it captures ~70% of the oracle's mean gain on Detroit but degrades Chicago. The prototype's internal confidence (inlier count, RMS) does not discriminate its dense-grid drift errors; I also tested an association-ambiguity margin (distance gap to the nearest alternative same-name interpretation per inlier) and it **fails to separate** harmful from helpful pages (~500–600 ft on both). The defensible rule today is **S1 gap-fill**: strictly additive coverage, never touches an existing fit.

## Limitations learned (the point of the exercise)

1. **Point-to-line-only least squares is degenerate.** A few labels in two direction families leave a grow-and-rotate solution family (first p85 solve: scale 0.66 → 2.34 ft/px in one step). Label-direction rows — conveniently linear in the transform — plus a scale clamp fix it. Any production version needs both.
2. **The region-adjusted scale prior doubles as a lie detector.** Every verified-but-wrong fit sat >20% off the prior (166, 243 ft — wrong parallel-branch lock-ins); everything within 10% measured ≤74 ft. Tightening the verified band to [0.85, 1.2] eliminated all bad fits at zero cost.
3. **Modern centerlines missing 1929 streets starve association entirely.** On Detroit p5 (industrial riverfront, streets erased by redevelopment), labels sit 400–1350 ft from their streets' surviving geometry *even under a correct transform*. Since the project's goal is to *produce* historical street datasets, historical centerlines will not exist to fix this — coverage on these pages will need **shape matching against the key-map region polygon** (page ink footprint ↔ georeferenced block outline) rather than street names.
4. **The nofit tail is evidence-starved, not fitter-limited.** Most pages the pipeline can't place have unusable labels: waterfront non-streets (p50: PARK, MOTOR, BOAT), names ambiguous in the index (RIVERSIDE ×4 variants), or ≤2 matchable street names. The prototype's honest role is tail-quality + split panels + conservative gap-fill, not a coverage machine.
5. **Selection is the open problem.** Per-page choice between two verified fits needs a signal neither fitter currently produces; the naive ambiguity margin doesn't work. Candidate directions: residual distributions under *both* transforms on the union of evidence (each fitter scored on the other's GCPs), or agreement/disagreement between the two transforms as its own signal (they agree within ~10 ft on most pages; large disagreement flags exactly the pages where one is wrong).

## Files

- `mapsnap/predict_fit.py` — prototype fitter + CLI (`uv run python -m mapsnap.predict_fit DIR --keymap ... --out DIR [--truth main.iiif.json]`), writes georef.json-compatible outputs, a per-page table, and `summary.json`.
- `mapsnap/predict_fit_test.py` — core-math tests (transform conventions, rotation voting, WLS recovery of a known transform).


danvk:
## Follow-up: shape-matching exploration (key-map regions × page ink footprints)

Can the pages with few/no usable street names be placed by matching the page's colored ink footprint against its georeferenced key-map region polygon? Explored on Detroit 1929 v11 (91 truth-scored unsplit pages; 14 of them are the "hard set" with no pipeline fit).

**Method**: ink mask from a downscaled page (dark linework ∪ saturated color, density-filtered to drop the border frame, largest component); scale from the volume prior (median × region multiplier); rotation from the volume's median fitted rotation with quarter-turn candidates scored by shape overlap; translation by anchoring the ink centroid to the region centroid minus a volume-calibrated systematic offset.

**Calibration first** (69 well-fit pages, truth-projected ink vs region): region centroids sit a median **285 ft** from the true ink centroid (196 ft after removing the volume-wide mean offset, but p75 stays ~350); regions are drawn **0.70× the true linear size**; principal-axis orientation is useless (median 34° error). That bounds everything downstream.

**Results (anchored fit)**:

| | ≤100 ft | ≤200 ft | ≤300 ft | median |
|---|---|---|---|---|
| all 91 truth pages | 7 (8%) | 31 (34%) | 41 (45%) | 351 ft |
| hard set (14) | **0** | **2** (p22 141, p30 156) | 5 | ~380 ft |

**Refining translation against the region outline makes it worse** (median 351→374, ≤200 ft: 31→17): the drawn outlines are schematic — the centroid is the only informative statistic, and it carries ±200–350 ft of noise. Two structural failures in the hard set: p6 (its key-map number is detected twice — a misread block elsewhere; shape overlap can't disambiguate, neighbor-tiling could) and p50 (waterfront block mislocated on the key map, 1784 ft).

**Answer to the headline question**: with regions as drawn, shape matching does **not** reach 100 ft on any hard page and reaches 200 ft on only ~2 of 14. What it does deliver is *approximate* placements (150–900 ft) for pages that today have nothing — useful as flagged/approximate coverage, debugger seeds, or manual-georeferencing starting points, in the same accuracy class as the earlier anchor-placement experiments.

**What could raise the ceiling**:
1. **Neighbor tiling** — pages tile the volume; jointly optimizing all shape anchors against adjacency (page N borders N±1) should cancel much of the per-region centroid noise, and disambiguates duplicate-number anchors like p6.
2. **Hybrid with predict-then-verify** — one matchable street label + the shape anchor over-determines the fit; the hard pages with 1–2 labels (p93-class) are exactly where the two evidence sources compose.
3. **Page-edge continuity** — adjacent pages share street geometry at their edges; matching ink structure across page boundaries is name-free and survives demolished grids.

Scripts in session scratchpad (`shape_calibrate.py`, `shape_fit.py`); say the word and I'll materialize them into the repo.

danvk:
## Follow-up 2: oracle page numbers + page-edge ink continuity (Detroit hard set)

**Q: with perfect page-number detection, how well does region-based placement do?**
Simulated with the hand-labeled key-map truth (`data/keymaps/detroit-p0.labels.json`, same 6447×7795 sheet) as the detector, with truth-seeded region segmentation (101/101 regions). Answer: perfect numbers **eliminate catastrophic anchors but do not add precision**. p6 (whose CNN anchor was a duplicate-number misread) goes 9485 → 342 ft; every other hard page moves marginally, because the binding constraint is the drawn regions' centroid noise (±200–350 ft), not the numbers. Two incidental methodology findings: letting region-shape overlap choose among quarter-turn rotations is actively harmful (fixing rotation to the volume median took p28 from 827 to 64 ft); and truth-seeded segmentation reshuffles individual regions (p52: 248 → 890 ft), i.e. segmentation noise is on the same order as detection noise.

Best region-only configuration (oracle numbers, volume-median rotation, corrected-centroid anchor), hard set of 14:

| | ≤100 ft | ≤200 ft | ≤300 ft | median |
|---|---|---|---|---|
| region anchor only | 1 (p28: 64) | 6 | 7 | 305 ft |

**Page-edge ink continuity.** Streets are the gaps between colored blocks, and the gap pattern along a page edge must continue into its neighbor. Prototype: normalized ink-density profiles (30 ft bins, 120 ft band) along each page's content edges in a volume-rotation-aligned frame; hard pages grid-search a translation maximizing profile correlation with opposing edges of already-fitted neighbors (1–4 seams/page).

The signal is real — it produced the two best hard-page placements of any method: **p62: 151 → 35 ft** (tight search) and **p50: 405 → 146 ft** (wide search; the demolished-waterfront page with a distinctive shoreline edge). But blind always-apply refinement is net-neutral: two failure modes tug against the search radius from opposite sides.

- **Lattice aliasing**: street grids are periodic, so a wide search (±480 ft) snaps to the wrong block multiple (p8: 388 → 988 ft, shift 738 ft ≈ 2 blocks).
- **Radius starvation**: a tight anti-aliasing search (±210 ft) can't reach true offsets beyond it (p50's ~300 ft, so 405 → 564).
- With only 1–3 seams of 12–40 bins each, spurious correlation maxima are common; there's no per-seam confidence yet.

| hard set (14) | ≤100 ft | ≤200 ft | median |
|---|---|---|---|
| anchor only | 1 | 6 | 305 ft |
| + edge refine (wide) | 0 | 4 | 299 ft |
| + edge refine (tight) | 2 | 4 | 304 ft |

**Where this points.** Per-page greedy refinement wastes the signal. The production shape is *joint*: optimize all hard pages simultaneously against all seams (including seams between hard pages), with per-seam confidence weights and the region anchor as the lattice-phase prior that suppresses aliasing globally — i.e., the neighbor-tiling idea and edge profiles are one optimization, not two features. The 35 ft result says the payoff ceiling is high; scripts in session scratchpad (`shape_calibrate.py`, `shape_fit.py`, `edge_refine.py`).

danvk:
## Follow-up 3: joint optimization of edge-continuity seams

Explored solving all hard pages simultaneously: per-page translation offsets from the region anchors, objective = Σ per-seam edge-profile correlations (including **hard↔hard seams** — the hard pages cluster on the waterfront, so p6/p7/p8, p28/p29/p30, p61/p62 constrain each other) minus a calibrated anchor prior ((|d|/300 ft)², the measured region-centroid noise), coordinate-descent sweeps. Four variants: raw correlation, background-subtracted correlation, plus two greedy baselines from follow-up 2, plus a refuse-first seam-voting scheme (each seam independently nominates a shift with confidence; apply only when ≥2 seams agree).

**Results (Detroit hard set, 14 pages, oracle page numbers):**

| variant | ≤100 ft | ≤200 ft | ≤300 ft | median |
|---|---|---|---|---|
| region anchor only | 1 | 6 | 7 | 305 |
| joint, raw correlation | **3** | 4 | 8 | 282 |
| joint, background-subtracted | 2 | 5 | 6 | 318 |
| seam voting (refuse-first) | 1 | 4 | 8 | 277 |
| *oracle: best variant per page* | *5* | *7* | *~9* | *~230* |

**The mechanism works.** The hard↔hard coupling produced the two best placements of the entire shape-matching exploration: **p61: 135 → 21 ft and p62: 151 → 37 ft** (raw-correlation variant, 4–5 mutually-consistent seams each), and p50's shoreline seam again reached 146 ft. When seams are genuine, joint edge-continuity places pages at street-fitting accuracy *with zero street names*.

**But no variant dominates**, and the failure is diagnostic rather than architectural:

- Different objectives pick different optima per page (p61 is 21 ft in one variant, 378 in another; p28's excellent 64 ft anchor gets dragged to 196–263 by spurious seams); coordinate descent oscillates between sweeps.
- The refuse-first voting gate — the correct design under this pipeline's philosophy — **never finds two agreeing confident seams** at current feature fidelity, and its single-strong-vote fallback is wrong more often than right.
- Per-page best-of-variants shows the information is present (5/14 within 100 ft, median ~230), i.e. the seams *contain* sub-100 ft alignment for a third of the hard set — the scoring just cannot identify which optimum is real.

**Bottleneck identified: seam feature fidelity, not the optimization.** All variants share 1-D ink-density profiles in 30 ft bins computed from 400 px thumbnails — near the sampling limit for 60 ft streets, so genuine and spurious correlation peaks are the same height. The next investment is full-resolution 2-D edge-strip matching (≈8 ft bins, actual street-gap geometry): sharper genuine peaks should let the two-seams-agree gate fire, at which point the voting scheme composes cleanly with the region anchors and, per the earlier follow-ups, would take the hard set to roughly: every page placed ≤~350 ft (anchors), a third of them ≤100 ft (validated seams), zero catastrophes (oracle numbers + refuse-first gates).

Scripts: `joint_refine.py`, `vote_refine.py` alongside the earlier ones in the session scratchpad.

danvk:
## Follow-up 4: trimmed pages, key-map adjacency seam graph, high-res profiles

Implemented the two suggested improvements plus the fidelity fix from follow-up 3, on Detroit's hard set (oracle page numbers throughout):

- **3% edge trim** before ink extraction (binding/scan bands). Verdict: **keep — the cheapest real win of the round.** Anchor placement improves overall: median 305 → ~230 ft, ≤300 ft goes 7 → 9 of 14.
- **Key-map region adjacency as the seam graph.** Correct and necessary infrastructure, with a quantified catch: the drawn regions are so undersized (0.70× linear) that adjacent pages' regions don't touch — a fixed 75 ft buffer finds only 46 edges over 100 regions. Dilating each region by **20% of its own size** recovers a proper tiling graph (194 edges). Edge-to-edge matching "where it applies" is now derived from the key map itself rather than guessed from anchor positions.
- **High-res profiles** (10 ft bins from 1100 px thumbnails, was 30 ft from 400 px).

**What the sharper seams revealed.** The two-seams-agree gate still essentially never fires — but now for a diagnosable reason: confident votes **bifurcate into a near-zero peak and a one-block-out alias** (p61: [−10, +220] ft; p62: [+10, +320] ft; p8: [−340, −230] ft). The near-zero votes are the true signal — and they say the anchor was already right. Applied as corrections the seams keep underperforming (voting median 239→259, joint 239→285, with individual wins like p16 337→239 and p85 197→162 cancelled by losses); interpreted as **verification**, they work: "some seam votes ≈ 0" is exactly the anchor-confirmation flag a production system needs to mark which approximate placements are trustworthy.

**New error source isolated.** Comparing anchors across preprocessing variants shows large per-page instability (p22: 105 → 405 → 427 ft; p28: 64 → 166 → 137 ft) — a big share of the ±200–350 ft "region noise" measured in follow-up 2 is actually **ink-centroid estimation variance**, not map-drawing error. That reframes the next lever: not better seam scoring, but a better page-footprint estimator — robust partial-shape registration of the region polygon onto the ink outline (tolerant of the region's systematic undersizing), or an ensemble anchor over preprocessing variants.

**Where the shape-matching arc stands after four rounds** (Detroit hard set, 14 pages): every page gets an approximate placement, best rounds put ~9/14 within 300 ft and the all-rounds best-per-page union reaches ~7–8 within 200 ft — but no single configuration achieves that union, and seams are better spent verifying anchors than moving them. The composed production picture: region anchor (+trim) for placement, adjacency-graph seams for confirmation flags, street-based fitters for precision wherever even one label survives.

PR #108, opened 2026-07-07
Mapsnap keymap

danvk:
Makes key-map processing fully automatic and both volume pipelines resumable. Previously a volume's key map had to be manually identified, downloaded at full resolution, and processed by hand, and `ocr`/`georef` had to be told about it with `--keymap`. This PR adds a command to build a key map's sidecars, a detector that finds which page(s) are key maps, wires both into the LOC and OIM pipelines with no manual step, fixes two bugs a real `run-oim` uncovered, and makes both pipelines skip already-completed work on a re-run.

## `mapsnap keymap <raw/pN.jpg>`

Runs the three sidecar steps on a full-resolution key map:

1. **OCR + georef** the key map from its own street labels (`--geocode_keymaps`, so georef doesn't skip it), at a key-map-appropriate `--min-short-side` (default 60 ≈ 4× the 25%-scale page default of 15, since the raw sheet is ~4× the linear resolution). Native-resolution tiling is already on by default, which is what makes the sheet's small labels detectable. → `<stem>.streets.json`, `<stem>.georef.json`
2. **CNN + CRNN page numbers** with `--pages` **auto-derived** from the volume's page images (e.g. `1-32,34-85`, correctly excluding missing pages). Passing `--pages` is what enables the narrow-detection re-read — there is no separate `--reread` flag. → `<stem>.keymap.json`
3. **Region segmentation** (`page_regions`). → `<stem>.regions.panels.json`

## `mapsnap keymap-detect <volume>` — which page is the key map?

The convention is messy (`p0`, split `p0L`/`p0R`, `p1b` among cover/title/index siblings, `p1` in small volumes), and no manifest metadata reliably flags it — LOC labels every page "Page N", and OIM excludes the key map from `main.iiif.json`. So this uses a content signal.

**Key finding: identify by page-set _coverage_, not detection _count_.** A key map is an index of the whole volume, so its numbers reconstruct the volume's own page set. Raw CNN candidate count does **not** separate key maps from regular pages — a dense downtown page can have *more* number-like glyphs (measured: detroit's regular p66 had 209 candidates vs the key map's 116). But coverage does, with a huge margin:

| volume | key map coverage | best regular page |
|---|---|---|
| chicago | 0.88 | 0.01 |
| washington | 0.93 (→ `p1b`) | 0.02 |
| champaign | 0.96 | 0.04 |
| detroit | 0.95 | 0.00 (p66, despite 209 candidates) |

Two stages: **Tier B** nominates the page-0/page-1 families from filenames (cheap); **Tier C** confirms by reading each candidate's numbers (CNN+CRNN, snapped to the valid page set) and keeping those with coverage ≥ 0.3. Runs on the always-present **25%-scale `p*.jpg`** — that's the CNN's native working resolution (`keymap_patches.working_scale` uses it as-is), so no full-res download is needed to decide what to fetch. Deliberately does **not** use OIM's `key.iiif.json` for identification (truth data must not be a pipeline input).

## Pipeline wiring + `ocr`/`georef` auto-discovery

Both pipelines call `identify_keymaps` after building centerlines, then `mapsnap keymap`:

- **LOC** downloads the identified key map(s) at full resolution via `mapsnap download-raw`.
- **OIM** already has `raw/`, so it **deletes the non-key-map raw images** (they duplicate the 25% `pN.jpg`) unless `--keep_raw` — a new default-on behavior for disk savings.

**`ocr` and `georef` now use key maps by default** (behavior change). When `--keymap` is not given, `resolve_keymaps`/`discover_keymaps` (in `keymap/locate.py`) find `<stem>.keymap.json` files — with a sibling `.georef.json`, so a failed georef is skipped — in the images' directory or under `raw/`. Precedence: `--ignore-keymap` (new) turns it off, an explicit `--keymap` still wins, otherwise auto-discover. `mapsnap keymap`'s own internal `ocr`/`georef` pass `--ignore-keymap` so a re-run doesn't locate the key map against itself.

## Pipeline fixes (from a real `run-oim`)

- **OIM never downloaded the key map.** It lives only in `key.iiif.json`, never in `main.iiif.json`, so `download-oim` (run only on `main`) never fetched it → identify found nothing. Since `download-oim` reads *any* AnnotationPage's `items` by label, `pipeline_oim` now runs it on **both** files; the key map lands in `raw/`, is scaled, and is found by content — the pipeline never learns it came from `key.iiif.json`.
- **Malformed `mapsnap georef` crash.** `fit` made `--tag` a *flag*, but both pipelines called `mapsnap fit <dir> mapsnap`; the bare positional `"mapsnap"` fell through `parse_known_args` into the georef passthrough args (`… --centerlines … mapsnap` → *"unrecognized arguments: mapsnap"*). Fixed to `mapsnap fit <dir> --tag mapsnap` in both pipelines.

## Resumable pipelines

Both `run-loc` and `run-oim` now resume: a re-run skips steps that already completed, so an interrupted or crashed pipeline picks up where it stopped instead of redoing downloads, scaling, OSM, OCR, etc.

A new `Step` context manager (`utils.py`) is constructed once with the directory and a `force` flag, then each stage is `with step("name"):` wrapping the existing `run_cmd` calls — no per-step `if/else`. Completion is recorded as an empty `<dir>/.pipeline/<name>.done` stamp written only on clean exit (`run_cmd` exits the process on failure, so an interrupted step leaves no stamp and re-runs). `--force` re-runs everything; OCR additionally gets `--resume` so a partway-interrupted OCR resumes per page within its re-run.

Skipping an already-completed step's `with`-body — which Python doesn't support directly — installs a line trace on the calling frame that raises the instant the body starts and swallows it in `__exit__` (the standard `sys.settrace` idiom, CPython-only), restoring the prior trace so an outer coverage/debug tracer keeps working. Verified on 3.13; the repo runs no coverage to collide with it.

## Testing

Unit tests for the new pure logic — coverage thresholds, candidate generation, page-spec derivation, discovery precedence / `.georef.json` filtering, the destructive `delete_other_raw`, and the `Step` skip/run/force/raise paths (including a chained skip-then-run resume). Validated end-to-end: `mapsnap keymap` reproduces chicago's hand-built sidecars (identical georef corners, 103 detections, 102 regions); `keymap-detect` returns the correct page on chicago/detroit/champaign and disambiguates washington's `p1a/b/c/d` → `p1b`; a real `georef` run auto-discovers the key map with no flags; and a `download-oim --dry-run` on `key.iiif.json` confirms it resolves to `raw/p0.jpg`. Full suite green (501 tests).

## Notes for the reviewer

- The scaled top-level key map (`p0.jpg`) is in `list_pages`, so the main run also processes it as a regular page. This is pre-existing behavior, but note it now actually triggers on OIM volumes since the key map is finally downloaded into the volume.
- The `Step` skip uses `sys.settrace`; it only touches tracing on the resume/skip path and restores it, but it is CPython-specific — flagged in case that matters.
- Argparse help strings need `%%` not `%` (a literal `25%` in `--keep_raw` help crashed `--help`; fixed).

### DIFF
--- a/mapsnap/cli.py
+++ b/mapsnap/cli.py
@@ -31,6 +31,10 @@ SUBCOMMANDS: dict[str, tuple[str, str]] = {
         "mapsnap.keymap.pipeline",
         "Prepare key map(s): OCR+georef, page-number detection, and region segmentation",
     ),
+    "keymap-detect": (
+        "mapsnap.keymap.identify",
+        "Identify which page(s) of a volume are key maps (from 25%-scale images)",
+    ),
     "iiif": (
         "mapsnap.make_iiif_georef",
         "Combine georeferences into a IIIF AnnotationPage",
--- a/mapsnap/cli.py
+++ b/mapsnap/cli.py
@@ -27,6 +27,10 @@ SUBCOMMANDS: dict[str, tuple[str, str]] = {
         "mapsnap.georef_from_labels",
         "Georeference a map from detected street labels",
     ),
+    "keymap": (
+        "mapsnap.keymap.pipeline",
+        "Prepare key map(s): OCR+georef, page-number detection, and region segmentation",
+    ),
     "iiif": (
         "mapsnap.make_iiif_georef",
         "Combine georeferences into a IIIF AnnotationPage",
--- a/mapsnap/detect_text.py
+++ b/mapsnap/detect_text.py
@@ -15,7 +15,7 @@ from PIL import Image
 from tqdm import tqdm
 
 from mapsnap.ctc_vocab_decode import HINT_STRINGS, generate_vocab_strings
-from mapsnap.keymap.locate import KeymapLocator, page_number
+from mapsnap.keymap.locate import KeymapLocator, page_number, resolve_keymaps
 from mapsnap.streets import build_block_index, polygon_side_lengths
 from mapsnap.utils import default_centerlines, image_stem
 
@@ -697,6 +697,15 @@ def main() -> None:
         ),
     )
     parser.add_argument(
+        "--ignore-keymap",
+        action="store_true",
+        help=(
+            "Do not use key maps. By default, when --keymap is not given, key-map detections "
+            "files (<stem>.keymap.json with a sibling .georef.json) next to the images or under "
+            "raw/ are discovered and used automatically; this flag turns that off."
+        ),
+    )
+    parser.add_argument(
         "--beam-width",
         type=int,
         default=20,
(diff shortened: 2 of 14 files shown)


danvk:
I ran:

    uv run mapsnap run-oim sanborn05511_036 data/hudson_co_nj_1950_vol_9 r957239 'https://s3.us-central-1.wasabisys.com/oldinsurancemaps/uploaded/documents/hudson_co_nj_1950_vol_9_'

It downloaded the keymap pages and detected them correctly, so this works well enough to merge.

PR #109, opened 2026-07-08
Make street matching space-insensitive

danvk:
Targets "WESTSIDE" vs. "WEST SIDE" in Hudson County, NJ 1950 vol 9.

This seems pretty targeted — it recovers 5/95 pages there and has no effect elsewhere. Claude flagged that "B ROAD" vs. "BROAD" is a possible failure mode.

### DIFF
--- a/mapsnap/streets.py
+++ b/mapsnap/streets.py
@@ -333,6 +333,17 @@ def canonical_street_match(
     return None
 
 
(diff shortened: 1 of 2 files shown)


danvk:
This feels a little too targeted. We can recover these pages in some other way, either via neighbors or geometry matching.

Issue #10, opened 2026-05-15
Figure out where images are by looking at the numbers of neighboring images


danvk:
The 425 in the top left is the page number for _this_ image, but the 473 on the top and the 426 on the right are numbers of neighboring pages. If we don't have any idea where this image is, but we do know where those images are, those numbers give us a much better sense for the placement.

<img width="2048" height="2928" alt="Image" src="https://github.com/user-attachments/assets/b016bd47-d121-4a2d-a0f3-f294ed11600f" />

danvk:
Largely solved by #110; I don't actually use this information for anything in the pipeline yet.

PR #110, opened 2026-07-08
Build page adjacency graph from numbers on page margins

danvk:
Sanborn sheets print the adjoining sheet's number on the margin (the key legend's "reference to adjoining sheet") — a big numeral across the shared boundary road, on the side that sheet continues. This PR reads those references and builds a per-volume **page adjacency graph**: an evidence source that is independent of the key map, works even where street names/geometry have changed since the OSM era, and — because each detection carries which edge of the image it sits on — pins a page's *orientation* from even a single detected neighbor. It's the first building block for pose-graph georeferencing (place the ~1/3 of pages that street matching can't, by anchoring them to neighbors that it can).

See #10

## `mapsnap adjacency <volume>`

Scans every non-split page image (skipping key-map sheets, whose faces are covered in page numbers; split *parents* are scanned since the printed references live on the full sheet's margins) with an EasyOCR digit pass, and writes a single `<volume>/adjacency.json`:

- **per page**: scanned image dimensions plus every digit read that parses to a valid volume page number — polygon, confidence, glyph height, position, image-relative edge (`T`/`B`/`L`/`R`, corners `TL`/`TR`/`BL`/`BR` — deliberately *not* compass terms, since page orientation is unknown at this stage), and nearest-OCR-box gap. Unfiltered reads are kept so filters can be re-tuned without re-OCR.
- **volume-level**: the `adjacency` list of mutual edges.

OCR runs as **one CRAFT detection + two recognition passes** sharing its boxes (detection dominates the cost and is allowlist-independent): a digits-only pass supplies the transcription — immune to look-alike substitutions like 0→O — and a letters-allowed pass over the same boxes vetoes the ones that are actually text (below). ~3.3 s/page on MPS (5 min for a 92-page volume); sharing detection is a 1.56× speedup over two full `readtext` calls, with byte-identical output.

## Claims, reciprocity — and why reciprocity is not enough

Raw digit reads are noisy (house/block/dimension numbers collide with valid page numbers), so a read only becomes a **claim** if it's a valid page number ≠ the sheet's own, near a page edge, and printed large; and a claim only becomes an **edge** when reciprocated (A claims B *and* B claims A). Directed claims are ~72% precise; mutual edges ~98%.

But the remaining 2% taught the real lesson: **reciprocity alone can't vouch for weak claims.** Junk claims are common enough to reciprocate by chance, and page numbering correlates with adjacency, so junk edges are often coincidentally "correct" — several edges this PR eliminates would have *passed* truth validation while feeding wrong orientations into a pose graph. Every filter below was validated by cropping actual pixels, not just by scoring against truth (all 52 rotated candidate claims turned out to be misread street names — "WESTSIDE", "Mc ADOO" — or `10" W Pipe` annotations; a conf-1.000 "8" turned out to be a digit torn off block number "1298"; a confident "55" turned out to be the street name "SMITH" force-transcribed under the digit allowlist). The resulting filter stack:

1. **Confidence ≥ 0.3** — measured on Hudson and Brooklyn, this floor loses no true edge on either volume while killing junk a permissive floor admits (forced-transcription street names read at ~0.1; genuine degraded references read above 0.3). 0.5 would kill one further junk edge but costs Hudson four real ones.
2. **Rotation ≤ 5°** — a rotated quad is never a sheet reference.
3. **Text veto (dual recognition pass)** — under a digits-only allowlist the recognizer is *forced* to transcribe street names as numbers; re-reading the same box with letters available lets text identify itself. A veto requires both a clear reveal (≥3 letters or letter-dominated) *and* the recognizer preferring that reading over the digit one — a genuine "41" also comes back letter-dominated ("AL", look-alike glyphs) but at lower confidence than its digit reading, while "SMITH" reads as text far more confidently than as "55".
4. **Single-digit confidence ≥ 0.9** — anything tall and narrow reads as a "1".
5. **Self-calibrated height band** — a two-pass scheme: provisional mutual edges → median height of their multi-digit claims (the printed size is remarkably consistent: Hudson median 46 px, p10–p90 of 42–48) → single-digit claims must fall within [0.65×, 1.4×] of it. Kills the 76–237 px frame-rule "1"s.
6. **Isolation floor (0.3× median)** — a genuine reference floats in whitespace; a digit fragment of a block number *touches* its sibling box (gap ≈ 0) while being a perfect glyph at the right size and confidence. This catches what confidence and size cannot.

Considered and rejected: reciprocal edge complementarity (B on A's right ⇒ A on B's left) — Hudson's diagonal grid makes only 101/158 *genuine* pairs complementary; blacklisting digits 1/8 — the isolation floor dominates it; and a naive letters-in-the-allowlist change — it kills genuine references whose digits flip to look-alike letters (four real edges across the two volumes), which is why the veto needs the confidence gate.

## Results

**Hudson Co. 1950 vol. 9** (`python -m mapsnap.score_adjacency`):

```
pages scanned:  92 (88 in >=1 mutual edge)
mutual edges:   165
precision:      165/165 truth-consistent (100%)
recall:         165/262 truth-adjacent pairs
```

**Brooklyn 1939 vol. 1**: 45 edges, 44/45 truth-consistent, 48/61 pages covered — one junk edge remains (p48–p9, numeric junk that reads as digits in both passes; a candidate for pose-graph loop-closure detection).

433 directed claims → 165 edges on hudson. The 30 pages that reach the graph despite failing street-based georeferencing include the whole waterfront/railyard cluster, connected in chains anchored by fitted pages — exactly the topology pose-graph placement needs. (Recall's denominator includes corner-to-corner contacts that sheets print no reference for, so it reads low.)

## Scoring: `python -m mapsnap.score_adjacency <volume>`

Evaluates an `adjacency.json` against OIM truth: two pages are truth-adjacent when their human-georeferenced footprints come within `--max-gap` (30 m) of each other. Reports precision/recall/coverage and lists contradicting edges. Its docstring carries the honest caveat from above: a truth-consistent edge can still rest on a coincidence-reciprocated junk read, so inspect survivors visually before trusting a new filter.

## Debugger: adjacency mode

Drop a page image + `adjacency.json` onto the debugger (the page is identified by the image's filename stem — e.g. `p49.jpg` → `p49`). The image overlay draws **mutual-neighbor claims in blue**, unreciprocated claims in amber, filtered reads grey-dashed; the side panel summarizes mutual neighbors with the edge each was printed on (`p40 (T), p50 (R), p51 (B), p52 (L), p54 (L)`) above a claims-first detection table with crop previews.

## Known gaps

- **Recall on cluttered margins is a detection problem, not a filter problem**: some genuine references are struck through by the red page-boundary stripe, pipe-line dashes, or the binding fold, and CRAFT misses or fragments them (Kansas City overprints references heavily; hudson p8's west "7" is unreadable). Red-stripe suppression before OCR is a candidate fix.
- Split sheets are scanned as whole parents; scanning panels separately might recover references for pages like p1 (currently uncovered).
- 4/92 hudson pages have no mutual edge (p1, p7, p85, p86 — degraded margins).

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -1,6 +1,7 @@
 import { useEffect, useMemo, useRef, useState } from 'react';
 import './styles.css';
 import type {
+  AdjacencyData,
   Corners,
   GcpPairResult,
   GeorefData,
@@ -15,11 +16,17 @@ import {
   directionThroughCorners,
   projectThroughCorners,
 } from './geometry';
-import { filterDetections, type DetectionFilters } from './detections';
-import { parseDroppedJson } from './fileLoading';
+import {
+  detectionFromAdjacency,
+  filterDetections,
+  type DetectionFilters,
+  type IndexedDetection,
+} from './detections';
+import { pageStem, parseDroppedJson } from './fileLoading';
 import { ImageColumn, type Mode } from './components/ImageColumn';
 import { MapView } from './components/MapView';
 import { GcpControls, type GcpFitStats } from './components/GcpControls';
+import { AdjacencyTable } from './components/AdjacencyTable';
 import { DetectionsTable } from './components/DetectionsTable';
 import { PanelsTable } from './components/PanelsTable';
 import { loadImage } from './loadImage';
@@ -116,6 +123,12 @@ export function App() {
   const [panelLabels, setPanelLabels] = useState<string[] | undefined>(
     undefined,
   );
+  const [adjacencyData, setAdjacencyData] = useState<AdjacencyData | null>(
+    null,
+  );
+  // Stem of the loaded image's filename (e.g. "p49"), which identifies the page
+  // in volume-level files like adjacency.json.
+  const [imageStem, setImageStem] = useState<string | null>(null);
   const [selectedIndices, setSelectedIndices] = useState<Set<number>>(
     new Set(),
   );
@@ -236,6 +249,31 @@ export function App() {
     [detections, filters],
   );
 
+  // Adjacency mode: the loaded image's page entry (identified by filename stem)
+  // and its digit reads converted to the Detection shape the overlay renders.
+  const adjacencyPage =
+    mode === 'adjacency' && adjacencyData && imageStem
+      ? (adjacencyData.pages[imageStem] ?? null)
+      : null;
+  const adjacencyDetections = useMemo<IndexedDetection[]>(() => {
+    if (!adjacencyPage || !adjacencyData || !imageStem) return [];
+    // Page numbers of this page's reciprocated neighbors; claims of these render blue.
+    const mutualNumbers = new Set<number>();
+    for (const [a, b] of adjacencyData.adjacency) {
+      const other = a === imageStem ? b : b === imageStem ? a : null;
+      const number = other ? adjacencyData.pages[other]?.number : null;
+      if (number != null) mutualNumbers.add(number);
+    }
+    return adjacencyPage.detections.map((d, i) => ({
+      det: detectionFromAdjacency(d, mutualNumbers),
+      i,
+    }));
+  }, [adjacencyPage, adjacencyData, imageStem]);
+  // Polygon coordinates live in the scanned image's pixel space; prefer its
+  // recorded dimensions so a different-resolution image still lines up.
+  const overlayWidth = adjacencyPage?.width ?? jsonWidth;
+  const overlayHeight = adjacencyPage?.height ?? jsonHeight;
+
   // Parse georef JSON text and update streets/intersections/corners/dimensions.
   function applyGeorefJson(text: string): void {
     let data: GeorefData;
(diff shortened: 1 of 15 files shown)


danvk:
Usage:

```
$ uv run mapsnap adjacency data/brooklyn_ny_1939_vol_1
Scanning 61 pages (61 page numbers)...
... 61/61 [03:35<00:00,  3.53s/it]
Single-digit height band: [29, 62]px, isolation floor: 13px (from 100 confirmed multi-digit references).
Wrote data/brooklyn_ny_1939_vol_1/adjacency.json: 61 pages, 213 directed claims, 46 mutual edges.
uv run mapsnap adjacency data/brooklyn_ny_1939_vol_1  88.80s user 12.26s system 45% cpu 3:40.26 total

$ uv run python -m mapsnap.score_adjacency data/brooklyn_ny_1939_vol_1
pages scanned:  61 (49 in >=1 mutual edge)
mutual edges:   46
precision:      44/46 truth-consistent (96%)
  single-digit: 1/2
recall:         44/166 truth-adjacent pairs (denominator includes corner contacts sheets don't print)
edges contradicting truth:
  p15 <-> p45
  p48 <-> p9
```

PR #111, opened 2026-07-13
Improved region detection

danvk:
Follow-up to #87.

#87 established colored-block segmentation of a key map via plain k-means. That worked well on saturated maps but **under-segmented pale, aged key maps** (Hudson County): blocks came out empty or at a fraction of their true size, and there was no way to *measure* how good any of it was. This PR:

1. adds a **truth-based scoring harness** so segmentation quality is a number, not an eyeball;
2. reworks the segmentation to fix the pale-map failure (seeded-palette quantization + chroma-family merge + ink/area guards);
3. teaches the page-number reader to **re-read duplicate labels**; and
4. documents a **SAM-based alternative** that was explored head-to-head and then removed.

## Measuring quality first

Nothing here is defensible without a metric, so that came first:

- **`truth_regions.py`** projects the volume's OIM page footprints through the key map's own georeference and drops any footprint less than 50% inside the image (crucial for two-key-map volumes like Brooklyn), writing a `<stem>.truth.regions.panels.json`.
- **`score_regions.py`** matches predicted panels to truth by page-number label (splits associated greedily by greatest intersection) and reports **mean IoU**, **count ≥ 0.5**, **misses**, and **predicted self-overlap**. Self-overlap is a free nonsense-detector: page blocks tile the map, so any overlap between two *predicted* regions is impossible ink.

### 1. Seeded-palette quantization (now the default)

**Root cause of the pale-map failure:** a pale block's fill is close enough to the aged paper that k-means lumps them into one cluster; the "largest cluster is background" rule then flags that cluster as paper and the block's seed is dropped.

**Fix (`cluster_image_seeded`):** build the color palette *from the seeds* — the median blurred-Lab color in a window around each detected page number — drop palette colors within ΔE 3 of the dominant global (paper) center, and greedy-dedup the rest at ΔE **16** (coarse wins: finer dedup fragments a single fill across near-identical centers). Palette clusters can never be classified as background, so a pale block keeps its seed. Falls back to plain k-means when no palette color survives (an uncolored key map); `--global-kmeans` keeps the old behavior for ablation.

### 2. Chroma-family merge

**Problem:** within-block mottle, fading, and stains shift a fill's *lightness* by more than the palette spacing, so one block's pixels split across 2–3 palette clusters and the region comes out clipped.

**Fix (`merge_cluster_families`):** union foreground clusters whose centers are close in Lab **(a, b) chroma** (ΔE 12) **and** whose pixels **interleave** (mutual contact ≥ 10% of the smaller). The contact gate is what distinguishes mottle (interleaved within a block) from a palette's genuine light/dark variants of one hue (which touch only along block borders) — an ungated chroma merge regressed the colorful maps.

### 3. Ink-veto and area guards

The naive family merge could produce a region covering 99.8% of the sheet: chroma matching ignores lightness, so the near-black **ink** cluster merged into a pale-tint family, and the ink lattice connects the whole map into one component. Two guards: (a) a **min-lightness floor** so near-black clusters never join a family; (b) a **per-region area cap** (a multiple of median seed-spacing²) with a fallback to the seed's *unmerged* cluster when a family is genuinely map-spanning, so the region shrinks to something sane rather than vanishing.

### 4. Split key maps and a scoring footgun

`handle split keymaps more gracefully`, plus a fix to `score_regions`' default truth-path derivation that could silently score a prediction file **against itself** (perfect 1.000) for unconventionally-named inputs — it now derives from the page stem and hard-errors on self-comparison.

### 5. Page-number reader: duplicate re-read

A label appearing **twice** is either a genuine split page (its number is printed on both blocks — Champaign) or a **misread of a similar number** (62 and 63 both squished until the CRNN fires only the central "6"; 80 misread as the already-found 30). `resolve_duplicate_labels` re-reads every duplicated detection from two tight, minimum-width crops and relabels one only when **both crops agree on the same page that is still missing** from the detected set — filling a hole in the coverage rather than shuffling labels, and leaving genuine split-page duplicates untouched. Live examples: Hudson `'6' → '62'` and `'6' → '63'`, Detroit `'6' → '96'`.

## Performance (before → after)

Mean IoU against projected-truth footprints, and the number of pages scoring ≥ 0.5. "Before" is #87's plain k-means; "after" is this PR's full pipeline.

| Key map | Before (mean IoU, ≥0.5) | After (mean IoU, ≥0.5) | Note |
|---|---|---|---|
| **Hudson p0** (pale) | 0.333, 42/95 | **0.499, 62/95** | the target case; degenerate/tiny regions 22 → 1 |
| Brooklyn p0 (colorful) | 0.477 | 0.478 | parity |
| Brooklyn p0b (colorful) | 0.539 | 0.537 | parity |
| Chicago p0b | 0.416, 48 | 0.410, 50 | parity |

The Hudson gain accrued in stages: seeded palette 0.333 → 0.478 (tiny regions 22 → 1), chroma-family merge → 0.493 (per-page: p16 .48→.75, p28 .33→.67, p42 .48→.71, p48 .54→.76), ink/area guards → 0.499 (and Hudson p44 sane at 0.21% of the sheet instead of 99.8%). The headline is **a large win on the pale map that defeated plain k-means, with parity and no regressions on the colorful maps that already worked**.

## The SAM alternative we explored (and why k-means won)

Because k-means fundamentally needs the block fills to be separable from paper, the natural question was whether a learned segmenter that doesn't care about color would do better. So this branch first built a full **Segment Anything (SAM)** pipeline (`sam_regions.py`, since removed): each detected page number is a positive point prompt on its block, every *other* page number is a negative prompt, and a box prompt sized 1.25× the nearest-seed distance fences the mask; overlapping masks are resolved to the nearest claiming seed (watershed-like) and each region keeps only the connected component containing its own seed.

It worked, and on the hard pale map it clearly beat *plain* k-means:

| | Hudson p0 (mean IoU, ≥0.5) |
|---|---|
| Plain k-means (#87) | 0.333, 42/95 |
| SAM, point prompts only | 0.325 |
| **SAM, box + negatives + nearest-seed resolution** | **~0.474–0.485, 59/95** |
| **Improved k-means (this PR)** | **0.499, 62/95** |

But once the k-means path was fixed (seeded palette + chroma-family merge), it matched SAM on Hudson and then edged ahead (0.499/62 vs 0.485/59), while remaining clearly better on colorful maps (Brooklyn p0/p0b: k-means 0.477/0.539 vs SAM 0.409/0.469). Critically, the two approaches **miss the same pages** — both are seeded by the CNN+CRNN page-number detections, so the remaining ceiling is *detection recall*, not segmentation quality, and SAM buys nothing there.

Given near-identical accuracy and a shared ceiling, the operational cost decided it. SAM requires a **~375 MB ViT-B checkpoint** and GPU inference and is much slower per sheet; the k-means path needs no model download, runs on CPU, and is far faster. So SAM and its `segment-anything` dependency were removed (last commit); the recipe and its numbers are preserved in the commit history and project notes for anyone who wants to revisit it.

(SAM variants that were tried and rejected, for the record: winner-take-all overlap resolution by mask score gutted the losing block (0.236); using the page-number bounding box as the box prompt segmented the glyph instead of the block (0.152); morphological opening of SAM masks consistently cost ~0.01.)


danvk:
Here's Claude's explanation of "∆E:"

"ΔE 3" means a color difference of 3 units in the CIELAB color space — I verified it maps to real code, so here's what it concretely is in this PR.

**The unit.** ΔE ("Delta E," E for the German *Empfindung*, "sensation") is the standard measure of *perceived* distance between two colors. The code converts each pixel to CIELAB via skimage's `rgb2lab` ([page_regions.py:168](mapsnap/keymap/page_regions.py#L168)), which gives the true CIE scale (L from 0–100, a/b roughly −128–127), and then measures color differences as ordinary Euclidean distance in that space ([`np.linalg.norm`](mapsnap/keymap/page_regions.py#L257)). Euclidean distance in CIELAB *is* the CIE76 ΔE formula, so calling these thresholds "ΔE" is accurate.

CIELAB is built so that equal geometric distances correspond to roughly equal *perceived* color change — unlike RGB, where a fixed numeric step looks like a big shift in some regions and a tiny one in others. That's the whole point of working in Lab here: a threshold like "3" means about the same perceptual thing regardless of hue. The rough scale:

- **ΔE ≈ 1** — the smallest difference a trained eye can notice (a "just-noticeable difference").
- **ΔE ≈ 2–3** — a small difference; the usual "these are basically the same color" tolerance.
- **ΔE ≈ 10+** — obviously different colors.
- **ΔE ≈ 100** — black vs. white.

**So "ΔE 3" in context** ([`palette_paper_distance = 3.0`](mapsnap/keymap/page_regions.py#L134)): a seed color within ΔE 3 of the dominant paper color is treated as *the same color as paper*. That's the tight "basically identical" tolerance — the page number is printed on open paper, not on a colored block, so its color is dropped from the palette rather than seeding a bogus region.

The description uses two other thresholds on the same scale, and the contrast is the interesting part:

- **ΔE 16** for palette dedup ([`palette_dedup_distance`](mapsnap/keymap/page_regions.py#L132)) — deliberately *loose*: two palette colors within ΔE 16 are collapsed to one, so a single printed fill that k-means split into near-identical shades doesn't fragment into several regions.
- **ΔE 12** for the chroma-family merge ([`family_chroma_distance`](mapsnap/keymap/page_regions.py#L135)) — same magnitude but measured on the **a/b (hue/chroma) channels only**, dropping L ([page_regions.py:359](mapsnap/keymap/page_regions.py#L359)). That's why the PR text calls it "Lab (a,b) chroma" rather than a plain ΔE: it intentionally ignores lightness so a mottled or faded patch (same hue, different brightness) still counts as one block.

One honest caveat: CIE76 ΔE is perceptually *non-uniform* — it overstates differences among saturated blues and greens — so these numbers were tuned empirically on these key maps rather than pulled from a universal perceptual constant. The just-noticeable-difference scale above is the right intuition for their magnitudes, not a precise guarantee.

Issue #78, opened 2026-06-18
Allow smaller, higher-confidence detections


danvk:
#68 involves setting min-short-side, min-long-side and min-confidence across a volume. But sometimes this is too coarse. It might make sense to admit smaller labels _if_ they're extremely high confidence.

For Brooklyn, the automatically-calibrated min-short-side is 18.0. So maybe we could have a schedule like:

- short-side 18, confidence 0.150
- short-side 16, confidence 0.300
- short-side 14, confidence 0.600
- short-side 13, confidence 0.990

Motivating examples:

On Detroit p29 the good detections are large and high confidence: E VERNOR (124x28, 0.758), KERCHEVAL (141x29, 1.000). CRATE→CRANE is small (46x14 and marginal confidence, 0.152).

<img width="1306" height="654" alt="Image" src="https://github.com/user-attachments/assets/3badea36-3f46-4c2c-bd30-89993568b910" />

Whereas on Brooklyn vol 1 p28, the detections are small but very high confidence: W 9TH (48x14, 0.936), HENRY (54x16, 1.000), HICKS (52x14, 0.997), COLUMBIA (86x16, 1.000), MILL (40x14, 0.998).

<img width="1561" height="633" alt="Image" src="https://github.com/user-attachments/assets/ba928414-13a9-4698-baee-5a949a601623" />

PR #117, opened 2026-07-15
Withhold the confidence size discount from collinear-dominated street labels

danvk:
## Idea

`confidence_relaxed_threshold` lets a **high-confidence label in below the volume's size floor** (issue #78). That discount is what admits Nashville p9's bogus `3RD`. This PR withholds the discount from a relaxed label that a **strictly better collinear label** — larger *and* more confident, naming a different street — sits on top of: on one straight run of roadway, the small faint one is usually noise riding on the big one's line, and accepting it fabricates an intersection.

The rule **only ever withholds the discount**. It never overrules the gate: a label that clears `--min-short-side` and `--min-confidence` on its own evidence is admitted no matter what shares its line.

## The case

Nashville **p9** reads `3RD` at **16 px / 0.319**. The volume's auto floor is 18 px, but `confidence_relaxed_threshold(0.319, 0.15, 18, 12.6)` drops the requirement to **15.6 px** — so a 16 px detection squeaks in on confidence alone. It sits **4 px** off the same vertical as a `7TH` at **24 px / 0.952**:

| | conf | short | centre | dir |
|---|---|---|---|---|
| 7TH | 0.952 | 24 px | (238, 1002) | 1.571 |
| 3RD | 0.319 | 16 px | (242, 529) | 1.571 |

That `3RD` produced the page's **only** — and wrong — intersection (`GAY × 3RD`), and the resulting 1-GCP deferred fit landed **1489 ft** out with scale **−44%**. Withholding the discount leaves p9 with **0 GCPs**, so it honestly reports no fit instead of emitting a wildly wrong georeference.

image

## Why the scoping matters

The tempting premise — *a straight run of roadway carries one name, so collinear labels naming different streets cannot both be right* — is *false*, and an unscoped rule acts on it:

- **A street changes name mid-run.**
- **A compound label names two streets at once.** p9's own `6TH` is part of `CAPITOL DR. OR 6TH AV N.` — a single printed label offering both names.
- **Two halves of a multi-word name are collinear by construction.** `ST` and `JEAN` share a line because they *are* one label (Detroit p20, p51, p57; New Orleans p489).
- **An undetected split page** puts two halves on one page line while they are hundreds of metres apart on the ground (New Orleans 1896 p125).

Scoping to relaxed admissions sidesteps all of these, because in each case the labels stand on their own size. p9's `6TH` (20 px / 0.896) clears the 18 px floor unaided and is now kept; `3RD` (16 px) still goes. Promoted avenue letters bypass the size gate entirely, so they are never relaxed admissions and are likewise immune.

## Blast radius

Measured across nine volumes (~660 fitted pages) against **this filter disabled**, on top of merged #116. **Six pages move.** Every volume's aggregate is unchanged except Nashville:

| Volume | Fit | mean | median | ≤15 | ≤25 | >500 | max |
| new_orleans_la_1951_vol_5 | 103 | 47.4 | 12.3 | 70 | 88 | 4 | 891.1 | 
| detroit_mich_1929_vol_11 | 83 | 41.0 | 14.8 | 42 | 56 | 0 | 429.3 |
| champaign_ill_1915 | 27 | 12.4 | 9.9 | 23 | 26 | 0 | 61.5 |
| brooklyn_ny_1939_vol_1 | 55 | 80.2 | 9.6 | 40 | 41 | 2 | 1706.2 |
| chicago_il_1950_vol_1 | 100 | 52.0 | 10.1 | 69 → **70** | 83 | 1 | 2989.2 |
| grand_rapids_mi_1953_vol7 | 63 | 62.8 → 63.0 | 18.4 | 21 | 45 → 44 | 2 | 1149.0 |
| hudson_co_nj_1950_vol_9 | 59 | 19.7 → 19.8 | 13.1 → 13.2 | 32 → 31 | 39 | 0 | 92.6 |

**Nashville**, the volume this is for:

| | Fit | mean | median | ≤15 | ≤25 | >500 | max |
| without | 54 | 73.7 | 13.3 | 30 | 36 | 1 | 1489.0 |
| **with** | 53 | **47.0** | 13.2 | 30 | 36 | **0** | **359.3** |

Every page that moves:

```
nashville  p9    1489.0ft -> NO FIT    the intended case: a fabricated GCP becomes an honest no-fit
chicago    p24N    20.8ft -> 14.9ft    better
brooklyn   p3      12.2ft -> 13.6ft    worse by 1.4
hudson     p61     13.1ft -> 18.1ft    worse by 5.0
grand_rap  p810    23.2ft -> 34.5ft    worse by 11.3
nola_1896  p125    11.9ft -> 771.5ft   pre-existing: undetected split page (see below)
```

An earlier, unscoped version of this rule moved **11** pages. Scoping it to relaxed admissions returned five of them to their unfiltered values — including two apparent *wins* (Brooklyn p1 48.0→57.2, p13 48.1→55.6) that were the rule overruling the gate and getting lucky. Declining those bets is the point.

## Known limits

- **New Orleans 1896 p125** (11.9 → 771.5 ft) is not caused by this rule so much as exposed by it. The page is a **split that the splitter fails to detect**, so `NUNS` and `ST. JAMES` — two clean, correctly-read labels of streets that are genuinely parallel and 221 m apart — land 9 px from the same page line. Any fit here is fragile: it only "works" while nothing happens to combine streets from either side of the undetected split. `NUNS` is 20 px against this volume's 24 px floor, so it is a relaxed admission and still eligible. The fix belongs in the splitter.
- **The same-street guard compares `normalize_street` output**, so `GATES AV` (→ `GATES AVENUE`) and bare `GATES` (→ `GATES`) read as different streets and the guard misses them (Hudson p61). It should compare canonical street matches. This affects ~4% of eligible drops and is left for a follow-up.

## Tests

7 unit tests: drops the worse collinear label; keeps a street labelled twice; keeps genuinely parallel streets; keeps a smaller-but-more-confident label; **spares a label that clears the gate alone** (p9's `6TH`); **spares a promoted label**; and `needs_size_relaxation` boundaries. Suite green; ruff + pyright clean.

### DIFF
--- a/mapsnap/georef_from_labels.py
+++ b/mapsnap/georef_from_labels.py
@@ -317,6 +317,11 @@ DOUBLE_SCALE_BAND = (1.8, 2.2)
 # treated as buildings. Brown is a dark yellow/orange and shares the band.
 FILL_YELLOW_HUE_BAND = (40.0, 110.0)
 
+# Perpendicular offset (page px) within which two same-bearing labels count as sitting on one
+# line for drop_collinear_dominated. 20 px admits Nashville p9's 4 px 3RD/7TH offset while
+# keeping genuinely parallel streets (a block apart) separate.
+COLLINEAR_PERP_TOLERANCE_PX = 20.0
+
 
 def is_split_page(stem: str) -> bool:
     """Whether an image stem names one panel of a split sheet (p21__1-style suffix)."""
@@ -1990,10 +1995,12 @@ def _init_worker(
     rectangle_index: tuple[dict[str, list[Block]], float] | None = None,
     truth_by_page: dict[int, list[list[list[float]]]] | None = None,
     keep_labels_on_fill: bool = False,
+    collinear_perp_tolerance_px: float = COLLINEAR_PERP_TOLERANCE_PX,
 ) -> None:
     """Populate _worker_state once per worker process (or once in the main process)."""
     _worker_state.update(
         keep_labels_on_fill=keep_labels_on_fill,
+        collinear_perp_tolerance_px=collinear_perp_tolerance_px,
         block_index=block_index,
         cos_phi=cos_phi,
         centerlines_path=centerlines_path,
@@ -2069,6 +2076,7 @@ def _process_one_image(image_path: str) -> tuple[str, ProcessResult]:
             keymap=keymap,
             truth_polygons=truth_polygons,
             keep_labels_on_fill=_worker_state["keep_labels_on_fill"],
+            collinear_perp_tolerance_px=_worker_state["collinear_perp_tolerance_px"],
         )
 
     with _captured_page_log(log_path, _worker_state["debug"]):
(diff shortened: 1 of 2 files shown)


danvk:
This wound up being more narrowly-scoped than expected, but I think it makes good sense.

PR #118, opened 2026-07-15
Stop a bare street-name fragment from claiming every street it prefixes

danvk:
## The bug

Street matching accepted a label that was **any prefix** of a key, so a label naming **one word of a multi-word name** claimed every street starting with it.

- Brooklyn v1: a lone **`VAN`** matched **all eight** VAN\* streets — VAN BRUNT, VAN BUREN, VAN DAM, VAN DYKE, VAN SICKLEN, VAN SICLEN AV, VAN SICLEN CT, VAN SINDEREN AV.
- Detroit: a lone **`ST`** (which normalizes to SAINT) claimed **54** streets.
- Nashville: a lone **`REP`**, read off a building abbreviation, claimed **REP JOHN LEWIS WAY SOUTH and NORTH** — fabricating the GCPs behind a 4428 ft fit.

Note the asymmetry: only the **first** word is dangerous. `BRUNT` alone matches nothing, because no street *starts with* "BRUNT ".

This is an **interaction bug**, not one commit's fault. The prefix rule predates #18; #91 added per-word vocabulary so a CRAFT-split "VAN BRUNT" decodes to itself rather than snapping to an unrelated word (`VAN → IVAN`). #91 made `VAN` *decodable*; the old prefix rule then fanned it out to all eight.

## The rule

A label may omit the street type and quadrant ("MAGAZINE" → MAGAZINE STREET), so a prefix match is sound whenever **everything left over is type and/or direction words** (`_covers_whole_name`, applied in all three matchers). When the remainder holds part of the *name*, the label is a fragment.

**Fragments are judged on length.** A short fragment is an abbreviation or particle that recurs across a whole city and attaches to anything — `ST`, `AV`, `VAN`, `REP`, `NEW`, `PIER`. A fragment of **5+ letters** is distinctive, and a sheet printing it usually *is* naming that street under an older, shorter name than OSM now carries. Both cases are the same shape, and only length separates them:

| label | claims | verdict |
|---|---|---|
| `REP` (3) | REP JOHN LEWIS WAY | a building abbreviation on a 1957 sheet — **rejected** |
| `VAN` (3) | 8 VAN\* streets | half a name — **rejected** |
| `HARBOR` (6) | HARBOR ISLAND STREET | Detroit p99's only real reading — **accepted** |
| `CLYDE` (5) | CLYDE PARK AVENUE | Grand Rapids p815's only real reading — **accepted** |

`HARBOR` and `CLYDE` are each a confident, printed label with **no sibling word on the sheet** to assemble with, so rejecting them is unrecoverable. Without the length allowance this PR cost Detroit p99 (20 → 133 ft) and Grand Rapids p815 (17.7 ft → no fit).

**The count is of the label as printed**, not its normalization. `normalize_street` expands `ST` → SAINT (5 letters) and `AV` → AVENUE (6), so counting the expansion would hand the allowance to precisely the abbreviations the rule exists to stop — a bare `ST` would reclaim all 69 of Detroit's SAINT\* streets. Hence `printed_letters()`.

**Deliberately not an ambiguity test.** "REP" is wrong because it is *half a name*, whether the volume holds one REP\* street or eight. An ambiguity gate would also keep matches whose text happens to hit exactly one street — `PIER` on "PIER No 4", `SHIP` on "SHIP YARD" (both confirmed by eye).

## Commits

1. **`admit a faint partner into multi-word street assembly`** — assembly required *both* parts to clear `min_confidence`, so a faint second word left the first stranded, and a lone "VAN" is not inert. Brooklyn p19's `DYKE` scores **0.096** but sits **1 px** off collinear from a **1.000** `VAN`, and "VAN DYKE" *is* a street. One part may now fall to `WEAK_PARTNER_CONFIDENCE` (0.05); its partner must still clear `min_confidence`, so **two weak reads never vouch for each other**. `_assembled_detection` takes `max()` rather than `min()` of the parts: the assembly is corroborated by a confident anchor, collinear geometry and an exact vocabulary hit, so its strength is its **best** part — `min()` would re-sink "VAN DYKE" below the accept gate. p19 goes 12 → **15 GCPs**; across 21 volumes assemblies go **160 → 249**.
2. **`require a bare street label to name the whole street`** — the whole-name rule, plus the `STREET_TYPES` additions it needs (WAY, TRAIL, ALLEY, SQUARE, PLAZA, PIKE, …) so labels that *do* name the whole street still match ("PRINTERS" → PRINTERS ALLEY).
3. **`drop CRESCENT from the unabbreviated street types`** — CRESCENT is the case that list's own comment warns about. Detroit's CRESCENT AVENUE is named *Crescent*, so listing CRESCENT aliased it to a bare `AVENUE` that a 1.000-confidence `AV` label matched. It also widened `W` from 8 candidates to 10 and `S` from 28 to 32. Removing it changes no volume's numbers.
4. **`admit a long, distinctive label to prefix matching`** — the 5-letter allowance and `printed_letters()`.

## Measured

Across the six README volumes, of 4424 street-matching labels (conf ≥ 0.5) the rule **kills 371 outright and narrows 152**. Total candidate streets claimed falls **30575 → 18352 (40% fewer)**. What it kills is exactly the target class:

```
ST x169   AV x84   ST. x40   PIER x26   NEW x13   VAN x11   AV. x9   OLD x3
```

Fit results over eight volumes, against `main`:

| Volume | Fit | mean | median | ≤15 | ≤25 | >500 | max |
| new_orleans_la_1951_vol_5 | 103 | 47.4 | 12.3 | 70 | 88 | 4 | 891.1 |
| new_orleans_la_1896_vol_2 | 82 | 65.5 | 30.3 | 25 | 37 | 1 | 771.5 |
| detroit_mich_1929_vol_11 | 83 | 41.0 | 14.8 | 42 | 56 | 0 | 429.3 |
| grand_rapids_mi_1953_vol7 | 63 | 63.0 | 18.4 | 21 | 44 | 2 | 1149.0 |
| champaign_ill_1915 | 27 | 12.4 | 9.9 | 23 | 26 | 0 | 61.5 |
| brooklyn_ny_1939_vol_1 | 55 | 80.2 → 80.1 | 9.6 | 40 | 41 | 2 | 1706.2 |
| chicago_il_1950_vol_1 | 100 | 52.0 → **48.7** | 10.1 | 70 | 83 | 1 | 2989.2 |
| nashville_tn_1957_vol1a | 53 → **54** | 47.8 | 13.3 | 30 | 36 | 0 | 389.8 |

The first five are **identical to main**. Every page that moves:

```
nashville  p47     NO FIT -> 32.3ft    <- the gain
chicago    p50N   443.5ft -> 118.0ft
brooklyn   p5      69.8ft -> 63.9ft    (the assembly commit)
nashville  p1__2  110.1ft -> 61.3ft    | reference scale 2.5860 -> 2.7716 px/ft
nashville  p65     11.2ft -> 6.7ft     | after p47 joins the fit set (+7%);
nashville  p16    359.3ft -> 389.8ft   | these four are deferred/median-scale
nashville  p25    347.9ft -> 377.9ft   | pages riding the reference, each moving
nashville  p10    292.7ft -> 317.6ft   | +8-10%. Not the rule.
nashville  p56    279.2ft -> 303.4ft   |
```

**The gain is real and it is the REP class.** #116's colour filter catches 27 of Nashville's 30 confident REP reads, but **3 survive** — their backgrounds are yellow, and #116 spares yellow *by design*, because yellow is ambiguous between aged paper, a taped-on patch and a frame building. p47's REP sits at hue 101.7, squarely in the spared band; this PR refuses it and the page recovers from no fit to 32.3 ft.

So the two filters are **complementary rather than redundant**: #116 judges *where a label sits*, this judges *what a label can name*, and each catches REPs the other structurally cannot.

The rest of the case is insurance. `VAN` no longer fans out to eight streets and `ST` no longer to 54; the Brooklyn pages where `VAN` fanned out already fit, because RANSAC tolerated the ambiguity. This removes the mechanism rather than relying on RANSAC to keep absorbing it.

## Tests

8 new tests: 3 for assembly, 5 for matching (short fragments rejected; a long distinctive fragment admitted; the printed-vs-expanded count; the type-only remainder unaffected at any length; `printed_letters` punctuation handling). 591 passed; ruff + pyright clean.

### DIFF
--- a/mapsnap/georef_from_labels.py
+++ b/mapsnap/georef_from_labels.py
@@ -317,6 +317,13 @@ DOUBLE_SCALE_BAND = (1.8, 2.2)
 # treated as buildings. Brown is a dark yellow/orange and shares the band.
 FILL_YELLOW_HUE_BAND = (40.0, 110.0)
 
+# Confidence floor for the *weaker* half of an assembled multi-word street. A part this weak is
+# never trusted alone; it is admitted only when its partner clears min_confidence and the two
+# concatenate into a real street name (assemble_multiword_streets). 0.05 reaches Brooklyn p19's
+# 0.096 "DYKE" — collinear with a 1.000 "VAN", and "VAN DYKE" is a street — while staying above
+# the 0.00-0.03 noise floor that fills these pages.
+WEAK_PARTNER_CONFIDENCE = 0.05
+
 # Perpendicular offset (page px) within which two same-bearing labels count as sitting on one
 # line for drop_collinear_dominated. 20 px admits Nashville p9's 4 px 3RD/7TH offset while
 # keeping genuinely parallel streets (a block apart) separate.
@@ -1701,7 +1708,12 @@ def _assembled_detection(first: dict, second: dict, text: str) -> dict:
     return {
         "polygon": [[round(x), round(y)] for x, y in box],
         "text": text,
-        "confidence": min(first.get("confidence", 0.0), second.get("confidence", 0.0)),
+        # The assembly is a joint hypothesis corroborated by three independent signals: a
+        # confident anchor word, collinear/adjacent geometry, and an exact vocabulary hit. Its
+        # strength is therefore its best part, not its weakest — taking the min would throw away
+        # exactly the evidence that justifies trusting a faint partner (Brooklyn p19's 0.096
+        # "DYKE" beside a 1.000 "VAN": min() would re-sink "VAN DYKE" below the accept gate).
+        "confidence": max(first.get("confidence", 0.0), second.get("confidence", 0.0)),
         "angle": first.get("angle"),
         "dir_pix": round(first.get("dir_pix", 0.0) % math.pi, 4),
         "long_side": round(r1 - r0, 1),
@@ -1716,6 +1728,7 @@ def assemble_multiword_streets(
     perp_tolerance_px: float = 20.0,
     max_word_gap_frac: float = 0.7,
     min_confidence: float = 0.5,
+    weak_partner_confidence: float = WEAK_PARTNER_CONFIDENCE,
 ) -> tuple[list[dict], list[dict]]:
     """Combine adjacent single-word detections into a multi-word street name.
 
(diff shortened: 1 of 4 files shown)


danvk:
This winds up being surprisingly narrow since #116 already took most of the wins. Still, I feel much better knowing that mapsnap no longer does crazy prefix matching like "REP" -> "REP JOHN LEWIS WAY SOUTH".

Issue #122, opened 2026-07-16
Use adjacency to disambiguate 1-gcp rotation


danvk:
I've gone through a few variations on when to accept 1-gcp fits. The current approach is to accept them when they have neighbors that match the rotation (#49). But now that I have adjacency data (#110), the neighboring pages are a much better way to resolve +/-180° ambiguity. If a 1-gcp fit produces a georeference that matches the neighbor claims, then it's probably good.

danvk:
Measured this end to end on the 12 truth volumes. **Adjacency does resolve the ±180° ambiguity, and a key-map check detects it perfectly — but `snap` already rescues both flips and places them better than the corrected 1-GCP fit would.** Numbers below so the trade-off is on the record.

## Scope: how much is actually at stake

| | count |
|---|---|
| 1-GCP fits across the 12 truth volumes | 61 |
| accepted (rotation confirmed, #49) | 50 |
| **rejected → `.georef-1gcp.json`** | **11 (18%)** |
| rejected *and* scoreable against truth | 8 |

What the rejected ones would have scored had they been accepted, against the accepted population:

| | ≤25 ft | 25–200 ft | ≥200 ft | median |
|---|---|---|---|---|
| accepted (n=46) | 17 (37%) | 20 | 9 (20%) | 34 ft |
| **rejected (n=8)** | **1 (12%)** | 3 | **4 (50%)** | **150 ft** |

Accepting them wholesale is **+1 page under 25 ft against +4 disasters**, so the existing gate is doing real work — rejected fits are 2.5× more likely to be disasters and their median error is 4× worse.

## The errors are rotation, but not only rotation

Substituting the truth rotation into each fit (same scale, same anchor GCP, i.e. what the pipeline would have produced given the right angle):

| page | rotation error | anchor error | as fit | with true rotation |
|---|---|---|---|---|
| DC p132 | **−162.8°** | 26 ft | 2225 ft | **46 ft** |
| LA p1475 | **−168.0°** | 183 ft | 1634 ft | 187 ft |
| DC p151 | −16.6° | **291 ft** | 249 ft | 287 ft |
| DC p171 | −14.4° | 47 ft | 205 ft | 53 ft |
| DC p216 | −6.4° | 22 ft | 95 ft | **22 ft** |
| NO 1951 p438 | −3.9° | 33 ft | 41 ft | 63 ft |
| Hudson p4 | −2.7° | 11 ft | 26 ft | **23 ft** |
[...]

Issue #50, opened 2026-06-10
Use key maps for location and mask


danvk:
If you can locate a few images in a volume, then the "key map" tells you where the others are located:

<img width="1420" height="2067" alt="Image" src="https://github.com/user-attachments/assets/da842e36-26db-4beb-a11e-06c5aa656be1" />

The colored areas could also serve as a guide for masks (#32).

A quick test suggests that EasyOCR isn't very good at picking out the page numbers, but maybe some other approach would work better. See also #10.

danvk:
The summary below is Claude's.

My take:

- This is doable but definitely not trivial.
- I think the best approach to finding regions is color segmentation, maybe after removing all the black text/street line pixels.
- Claude says EasyOCR does fine recognizing the digits on full-scale key map images, but not the downscaled ones.
- There's considerable variation between volumes:
  - Champaign: the colored regions directly abut one another. There are identically-numbered, separated regions for the splits, e.g. 21 and 23.
  - Chicago: there are some taped-on areas on the key map that mean it has less consistent colors across single regions, and contrast is much worse.
  - Brooklyn vol 2 keymap 2: Good contrast, cleanly separated by streets.
- This would majorly complicate the pipeline by making it more circular. But the potential win from knowing roughly where each page is are huge.

## Findings from a prototype attempt (key-map → location)

I built and tested a full prototype of the key-map approach and iterated hard on recognition across three volumes (Brooklyn 1939 v2, Champaign 1915, Chicago 1950 v1). **Conclusion up front: the pipeline architecture works, but per-region digit recognition on these key maps is not yet reliable enough to be useful, and no single parameter set generalizes across key-map styles.** Parking it here for the future.

### What was confirmed about the OCR difficulty
- At the **scale the main pipeline uses** (pages downscaled to ~2048px for OCR), key-map numbers blur and merge into garbage. At **full resolution on a region crop**, EasyOCR reads bold numbers well. So the recognizer is fine *given a good crop* — the failure is scale + the main pipeline's street-vocabulary CTC constraint and `--min-long-side` filtering discarding small numerics. A dedicated full-res digit pass is required.
- A non-obvious, high-impact gotcha: **white-masking a crop outside the region polygon cripples CRAFT detection** (digits found in 11/30 masked crops vs **29/30 padded, unmasked**). The hard polygon edge confuses the detector. Use padded, unmasked crops and pick the right number by point-in-polygon instead.

### What the prototype does (end-to-end, on branch `split-opus`)
[...]

PR #125, opened 2026-07-17
Fit neighbors using road geometry

### LINKED ISSUE #10: Figure out where images are by looking at the numbers of neighboring images

danvk:
Fixes #10 
Relates to #122

Street-name OCR + RANSAC georeferencing tops out around 60–85% of pages per volume: the rest have too few readable street names, or streets that no longer match OpenStreetMap. This PR places many of those pages **without reading anything on them**, by matching the *road geometry* they share with their already-georeferenced neighbors along sheet boundaries.

Results on the four volumes with ground truth (pages placed / within 25 ft / 50 ft / 100 ft of truth):

| volume | RANSAC only | + neighbor fitting |
|---|---|---|
| washington_dc_1916_vol_2 | 79/125 — 37/60/71 | **106/125 — 47/80/95** |
| hudson_co_nj_1950_vol_9 (held out of model training) | 58/92 — 39/54/58 | **76/92 — 43/59/66** |
| detroit_mich_1929_vol_11 | 81/101 — 54/64/69 | **84/101 — 56/66/71** |
| brooklyn_ny_1939_vol_1 | 55/61 — 41/44/51 | 55/61 (unchanged — see limitations) |

4-dc-placement-map

*DC volume layout colored by placement source. Blue pages were fit by RANSAC; green and orange pages could not be fit from their own street names and are placed purely from neighbor geometry. The isolated blue rectangle at top right is a wrong RANSAC fit (1497 ft off) — the kind of error the pose graph detects and heals under its "replace" policy.*

<details>
<summary><b>Glossary</b> (click to expand)</summary>

- **GCP** — ground control point: a (page pixel ↔ longitude/latitude) correspondence. Street intersections matched to OSM give point GCPs.
- **RANSAC** — RANdom SAmple Consensus: the existing fitting algorithm that repeatedly fits a transform to random GCP subsets and keeps the one most other GCPs agree with ("inliers").
- **Effective GCPs** — the number of *physically distinct* intersections among a fit's inliers. OSM name variants (SOUTH CAPITOL ST SE / SW / bare) can expand one intersection into six inlier records; a "6-inlier" fit anchored on one point is really a 1-GCP fit, free to swing about it.
- **UNet** — a convolutional neural network shape for per-pixel prediction: an encoder halves resolution several times (context), a decoder scales back up with skip connections (precision). Here: grayscale page in, per-pixel P(road) out.
- **IoU** — intersection over union: overlap area ÷ combined area of predicted vs. expected road masks. 1.0 is perfect.
- **Seam / strip** — the band of duplicated content two adjacent sheets share along their common boundary street.
- **NCC** — normalized cross-correlation: a template-matching score, computed here over all translations at once via FFT, masked to valid page regions.
- **Chamfer distance** — for each predicted road point on one page, the distance to the nearest road on the other page. Minimizing it aligns the two road networks.
- **Verification score** — the matcher's truth-free confidence in a join: inlier fraction of skeleton points + fine-scale NCC − mean chamfer penalty, with disqualifiers.
- **Pose** — a page's placement: (x, y, rotation θ) at the fixed volume-median scale.
- **Pose graph** — an optimization where pages are nodes, relative measurements between adjacent pages are edges, and all poses are solved jointly instead of one at a time.
- **Huber loss** — a robust loss: quadratic for small errors, linear for large ones, so a few grossly wrong measurements get outvoted instead of dragging the solution.
- **EM** — expectation-maximization style alternation: solve poses holding per-edge candidate choices fixed, then re-choose each edge's candidate to best fit the poses, and repeat.
- **Line factor** — a constraint that a detected street label must lie *on* its named OSM centerline: it pins the page perpendicular to the street and its rotation, leaving the along-street position free.
- **IIIF** — International Image Interoperability Framework; a Georeference AnnotationPage is the standard JSON format for "this image maps to these coordinates."
</details>

## How it works

**1. Road model.** A UNet (`road_model.py`) turns each 25%-scale page scan into a P(road) heat map. Roads are the one feature that survives both Sanborn lithography and 100+ years of urban change, and the network suppresses the repeating block/lot texture that defeated earlier matching attempts.

1-road-model

*a DC page and its P(road) heat map.*

**2. Seam matcher** (`edge_join.py`, truth-free). For an anchor page with a known pose and an unplaced neighbor (adjacency comes from the printed neighbor-sheet numbers, #113):
- rotation candidates from road-orientation histograms plus the volume's *measured* set of neighbor-rotation offsets;
- translation by masked FFT NCC over blurred P(road) maps, restricted to an overlap annulus (sheets share a strip, never more than ~40% of a page);
- top-K peaks refined by chamfer least squares (continuous rotation, fixed volume-median scale — per-page free scale reliably made things worse);
- physical gates: both sheets print their neighbor's number on the shared side, giving two direction constraints; the key map (#50) gives an absolute position disc and a region polygon; candidates overlapping already-placed pages by more than a strip are impossible.

3-seam-contact

*two adjacent DC pages' road masks rendered at their true poses; the yellow band is the shared seam the matcher aligns.*

**3. Greedy chain** (`edge_join_experiment.py chain`). Verification-gated joins (≥ 1.2) from vetted anchors; newly placed pages anchor the next round; multi-anchor placements are fused. High precision, but it refuses weak evidence and can never revisit a mistake.

**4. Pose graph** (`edge_join_graph.py`, `posegraph` subcommand). The global solve that picks up what the chain refused:
- **every** mutual-adjacency edge contributes a relative-pose measurement — including sub-gate ones (verification ≥ 0.5), weighted by a sigma ladder; edges where *neither* page is fitted use synthetic keymap-posed anchors (the anchor's error cancels in the relative measurement), which lets unfitted islands stitch together;
- each edge carries its top-5 candidate poses; an EM loop plus a trial-switch pass (tentatively re-solve with each alternate on suspicious edges, keep it only if the global cost drops) lets the solver un-pick aliased winners — on a self-similar street grid the true pose is often ranked *second* behind a one-block slide;
- absolute evidence: RANSAC pose priors tiered by **effective GCPs** (1-point fits get 25 m sigmas — 10 of Detroit's 11 one-point fits are >50 ft wrong), key-map centroid priors, and **street-label line factors** (labels accepted by the exact same `prepare_label_features` gates as the main pipeline, at a higher confidence threshold; a 40 m clamp makes bad reads self-disable). Line factors run as a second stage inside a 15 m trust region of the seam solve, so they snap near-correct pages onto their streets but can't relocate a page onto a wrong same-name street;
- a robust (Huber) scipy least-squares solve over all (x, y, θ) at once; single-measurement pages must still meet the chain's 1.2 gate to be output.

**5. Hybrid + outputs.** `report --chain-source chain,posegraph` scores the combination (chain placement where the chain succeeded, graph elsewhere); `materialize` writes `pN.georef-neighbor.json` files the debugger renders; `mapsnap iiif` accepts comma-separated globs (`p*.georef.json,p*.georef-neighbor.json`) to build a whole-volume IIIF AnnotationPage viewable in the #123 viewer.

## Alternatives and dead ends

Earlier page-to-page matching attempts (before this branch) all failed on the same volumes: **SIFT keypoints** (the two sheets render the shared strip differently — dense on one, sketchy on the other), **raw-ink chamfer matching** (the regular block grid aliases: every offset by one block scores well), and **block-number OCR matching** (~25% recall at 25% scale). Road masks are the thing that made correlation viable.

Within this approach, the failure modes that shaped the design (each is a gate or mechanism above):
- Raw NCC's best pose slides the neighbor *onto* the anchor (maximum overlap = maximum correlation) → overlap annulus.
- Sliver contacts (corner-adjacent pages) overfit chamfer with perfect scores → minimum strip-point count, residuals restricted to the seam.
- 180° flips score as well as truth on a near-symmetric corridor, and a flip that also swaps sides defeats a one-sided check → *both* pages' printed-neighbor-number sides are enforced, plus the absolute key-map disc.
- One-block lattice slides beat truth by hundredths of a point while sitting 40% outside the page's key-map region and overlapping an already-placed neighbor → containment bonus + placed-page exclusion.
- In the pose graph: loose anchor priors let good RANSAC fits drift and poisoned everything downstream (tight tiers won); tiering by raw inlier count trusted a 206-ft-wrong "6-inlier" one-point fit (effective GCPs fixed this); using *only* the best fits as anchors backfired badly — on a self-similar grid, many weak pins beat few perfect ones, because unpinned clusters slide onto aliased lattices; greedy per-edge candidate re-assignment got stuck in basins where the graph had deformed to fit a wrong candidate (the trial-switch pass escapes, and there's a unit test reproducing it); street line factors applied without a trust region *hurt* pages that started >40 m wrong (wrong-street capture); and a naive "one name per label" ambiguity rule silently discarded DC's most common label type (K ST NE/NW/SE/SW are quadrants of one street, now merged into one factor).

Why keep both chain and graph? The chain is a precision filter with sequential context (it knows what's already placed); the graph is an integrator that uses weak evidence, closes loops, and revises mistakes. Today the hybrid beats either alone; per-page selection between them is future work.

## Example commands

Full pipeline for one volume (each step caches; re-runs are cheap):

```sh
uv run python -m mapsnap.edge_join_experiment infer data/washington_dc_1916_vol_2        # P(road) maps
uv run python -m mapsnap.edge_join_experiment chain data/washington_dc_1916_vol_2 --anchor-pose generated
uv run python -m mapsnap.edge_join_experiment posegraph data/washington_dc_1916_vol_2 --street-factors
uv run python -m mapsnap.edge_join_experiment report data/washington_dc_1916_vol_2 --chain-source chain,posegraph
uv run python -m mapsnap.edge_join_experiment materialize data/washington_dc_1916_vol_2 --source chain,posegraph
uv run mapsnap iiif data/washington_dc_1916_vol_2/main.iiif.json \
    "data/washington_dc_1916_vol_2/p*.georef.json,data/washington_dc_1916_vol_2/p*.georef-neighbor.json" \
    --output data/washington_dc_1916_vol_2/hybrid.iiif.json
```

With truth data present, `sanity` measures the mask ceiling (seam IoU at truth poses) and `report` prints the placement tables above. The hybrid IIIF file loads in the #123 viewer: `npm run iiif` + `npm run dev` in `app/`, then `?view=iiif&iiif=data/<vol>/hybrid.iiif.json`.

## How the road model was trained

Labels are free: for any page RANSAC already fit, rasterizing OSM centerlines through the fit's transform paints a road mask over the scan (`rasterize_road_mask`). No hand labeling.

```sh
uv run python -m mapsnap.train_road_unet \
    data/brooklyn_ny_1939_vol_1 data/champaign_ill_1915 data/chicago_il_1950_vol_1 \
    data/detroit_mich_1929_vol_11 data/grand_rapids_mi_1953_vol7 data/kansas_city_mo_1951_vol_4 \
    data/nashville_tn_1957_vol1a data/new_orleans_la_1896_vol_2 data/new_orleans_la_1951_vol_5 \
    data/san_francisco_calif_1950_vol_2 data/washington_dc_1916_vol_2 \
    --val data/hudson_co_nj_1950_vol_9 --base 32 --epochs 24
```

The checked-in `models/road_unet.pt` (v2) was trained on ~700 pages from those 11 volumes with three choices that mattered:
- **only well-fit pages** (effective GCPs ≥ 3) — a fragile fit paints roads in the wrong place, so bad fits poison the labels;
- **edge-emphasized sampling** — half of all training patches are pinned to the outer 320 px band, since the matcher reads the model exclusively in seam strips, where content is sketchier than the page interior; checkpoint selection uses a held-out *edge-band* IoU;
- **held-out validation on Hudson** — the honest generalization test.

Hudson (which the v1 4-volume model failed on — seam IoU 0.31, joins unusable) improved to 0.44; Detroit 0.46 → 0.57, Brooklyn 0.42 → 0.51.

2-hudson-v1-v2

*the same held-out Hudson page under both models; v2's corridors are cleaner and reach the page margins.*

## Limitations and next steps

- **Grid aliasing is not fully solved.** Two DC pages (p204/p205) still land one block off in the pure graph (~330 ft); the chain's sequential exclusion gate catches them, which is why the hybrid exists. Iterating measurement collection against graph-posed footprints would give the graph the same context.
- **Chain-first isn't always right.** The hybrid prefers the chain wherever it placed a page, but the graph's pose is sometimes better (Hudson p14: 412 ft chain vs 75 ft graph). A truth-free per-page selector — graph when its support is redundant and consistent, chain otherwise, refereed by evidence neither optimized (street-label agreement, key-map containment, overlap violations) — is the clearest next win.
- **Coverage is bounded by the adjacency graph.** Brooklyn's six unplaced pages have *no* mutual neighbor-number edges at all — no seam any matcher could read. Detroit's graph is similarly sparse. Better adjacency recovery (or key-map/street-only placement) is a separate axis from mask quality.
- **Isolated weak pages get rough placements.** Island clusters reachable only through weak gateways land at 150–700 ft (flagged in the output). Multi-round trust-region iteration of the street factors may walk them home.
- **Split pages** (pN__1, pN__2) are out of scope for the matcher in this version.
- **Fixed volume-median scale** trades a few feet on off-scale pages for robustness; a bounded per-page scale observation (not a free variable) could recover it.


danvk:
This is still an "experiment" but I'm going to merge it to avoid getting conflicts in the future, and so that I can get some experience using it. Inter-page fitting is clearly the path forward, and will likely be the headline feature for mapsnap 2.0.

Issue #128, opened 2026-07-17
Match on road geometries alone


danvk:
With #125, we have P(road) for each pixel. With a good scale and rotation prior, and a rough location prior, we can do a similar rendering of the OSM grid in the area and "snap" to the best location using IoU or correlation. This would produce a fit that's robust to street name changes or OCR mistakes.

For regular grids, there's a significant risk of being off by one block with this approach. So this might be best incorporated as a signal into a broader pose graph formulation.

danvk:
Here are a few examples of renamed streets that 🤖 Claude found in Detroit. These are interesting because they're only OCR'd because they happen to match streets many kilometers away. But for matching to OSM, that's useless.

> Measuring each dropped label's position (via the v1.2 fit) against the page's keymap circle: **every one sits 60–250 m from the center, against a 514 m radius** — and 0–80 m from the page's segmented region polygon. For example p56's MARSH is 97 m from center (417 m of slack).
> 
> What's outside the circle is the **OSM feature that donated the name to the vocabulary**. The nearest feature named "Marsh" in Detroit's 46k-feature centerlines is 44 km away; p5's GLADWIN/HOWE, p21's EWERS/LOMA, p32's whole set — all the same, 4–40 km. These are historic street names with no nearby modern OSM counterpart; in v1.2 they were vocab-legal reads that matched distant same-named streets, and tellingly, **almost every one was a RANSAC outlier** in the v1.2 fit. The restriction pruned junk, and pruned it correctly.

danvk:
🤖 Claude's initial exploration is promising.

> The road UNet opens up lots of possibilities beyond just edge matching. For example, https://github.com/danvk/mapsnap/issues/128 proposes using the road geometry itself to align pages with OSM, taking coarse fits (say from keymap and region priors) and upgrading them to tight fits. This could help in the Detroit examples you identified earlier where MARSH in the Sanborn map only matches a street many kilometers away. It could also be extremely useful in Washington, D.C. where EasyOCR still struggles with street names like "AVE O" and "AVE I". This will be highly susceptible to block aliasing in regular grids, but turning a continuous optimization into a discrete one (with just a few possible fits) could be a win on its own.
>
[...]

PR #152, opened 2026-07-23
Geometry-first georeferencing (osm-snap): rescue, arbitration, refinement — score 55.5 → 69.8

danvk:
Adds a **geometry-first georeferencing channel** that matches each page's road-UNet P(road) mask directly against OSM street geometry — no street-name OCR in the loop — and uses it three ways: to **rescue** pages the RANSAC georeferencer couldn't place, to **arbitrate** (replace) placed fits that OSM actively contradicts, and to **refine** mid-tier fits to street-grid precision. Validates the thesis of #128.

**Land-weighted net score across the 12 truth volumes: 55.5 → 69.8** (good share 58.7% → 72.7%, disaster share 3.1% → 2.9%). Every volume improves. Two volumes (NO-1951, Champaign) now exceed the 90-point target.

## Per-volume journey

Each stage's IIIF (plus its `mapsnap compare` `.txt` for the debugger) is checkpointed in every volume directory: **`snap-base`** (RANSAC only) → **`snap-union2`** (+ rescue & split panels) → **`snap-arb`** (+ arbitration) → **`snap-refine`** (+ refinement = the production configuration).

| Volume | base | +rescue | +arbitration | +refinement | total Δ |
|---|---|---|---|---|---|
| Washington DC | 36.6 | 58.8 | 58.8 | **76.1** | **+39.5** |
| Hudson | 33.5 | 37.8 | 39.9 | **49.5** | +16.0 |
| Nashville | 39.1 | 40.8 | 45.5 | **53.8** | +14.7 |
| Kansas City | 53.9 | 55.2 | 56.0 | **68.3** | +14.4 |
| New Orleans 1896 | 39.0 | 40.1 | 41.8 | **49.7** | +10.7 |
| Los Angeles | 66.5 | 70.0 | 70.8 | **76.1** | +9.6 |
| Detroit | 59.4 | 62.6 | 63.7 | **68.3** | +8.9 |
| New Orleans 1951 | 84.4 | 84.4 | 85.5 | **93.3** | +8.9 |
| Grand Rapids | 57.9 | 57.9 | 57.9 | **64.0** | +6.1 |
| Brooklyn | 71.6 | 73.8 | 73.8 | **77.6** | +6.0 |
| Chicago | 69.4 | 69.4 | 69.4 | **75.2** | +5.8 |
| Champaign | 90.5 | 90.5 | 90.5 | **92.7** | +2.2 |
| **AGGREGATE** | **55.5** | **60.3** | **61.1** | **69.8** | **+14.3** |

(The pre-existing baseline was reported as 55.0 before this branch; +0.5 of the journey is a truth-accounting fix, not pipeline improvement — see "missing truth splits" below.)

## How it works

**Matching (`mapsnap/osm_snap.py`, truth-free).** OSM centerlines are rasterized into a local equirectangular metre frame (2 m/px, 12 m stroke — an OSM analog of the road-UNet's P(road) output). The page's P(road) map is matched against it with the edge-join machinery from #125: masked FFT cross-correlation proposes top-K translations per candidate rotation; chamfer least-squares on the road skeleton polishes each pose. Search is centered on the page's key-map location with a **per-volume calibrated radius** (p90 of fitted pages' keymap-vs-fit residuals — tightening Hudson from ~500 m to 150 m was one of the single biggest levers).

**Rotation & scale priors** seed the search as explicit, provenance-tagged ladders:
- *label-pair-exact*: two OCR'd labels on distinct streets fix the exact rotation (the street tangent gives it mod 180°; the world offset between the two streets resolves the flip);
- *label-osm-mod180*: one matched label, both flips emitted;
- *adjacency-keymap*: printed neighbor-number directions vs key-map region centroid bearings;
- *ransac-neighbor*: rotation borrowed from fitted neighbors (or a fitted sibling panel — the same physical sheet);
- *mask-mod90*: dominant road-orientation sweep, always appended so the set is never empty;
- scale: volume median, plus a half/double family rung when the key-map region area disagrees by ~2×.

**Ranking.** Each candidate carries the matcher's verification (chamfer inlier fraction + fine content correlation − mean chamfer) plus soft evidence: **street-name alignment** (labels projected through the pose, snapped to same-named OSM streets — a boost, never a gate, so renamed streets still georeference), key-map region containment, and rotation-prior agreement.

**Selection** (`osm_snap_experiment.py`, `select`) combines two dev-calibrated committees: a per-page argmax with score/margin gates (margin measured against the best *distinct* lock, not near-identical twins) and a volume-wide energy minimization (footprint overlap against fitted pages — with thresholds calibrated from how much the volume's own adjacent fitted sheets' footprints overlap; no page-to-page seam matching is involved — plus key-map adjacency-consistency and side terms, solved exactly per connected component). Split panels additionally face **sheet-agreement gates**: a panel's affine extends exactly to an implied placement of the whole scanned sheet (the crop is a pure translation), and candidates must agree with any reliable fitted sibling's implied sheet (effective GCPs ≥ 3, ~60 m tolerance); no-anchor panels need a much higher solo score. **Caveat, found in review**: measured against every fitted sibling pair in the corpus, real Sanborn split pages are *inset composites* (separate neighborhood maps on one sheet of paper — implied sheets disagree by 233–2143 m), not contiguous maps cut apart. So in practice this gate blanket-blocks rescue of panels that have a reliable fitted sibling rather than discriminating geometrically. That conservative behavior is exactly what the 12-volume scores validated (it is what stopped KC's four would-be panel disasters, including a 753 ft lock at select 2.58 that no score threshold would catch), but the mechanism's honest description is a block, not a constraint check — see next steps.

**Arbitration & refinement** run the same evidence head-to-head on placed fits: `evaluate_pose` scores the incumbent RANSAC pose with the identical verification formula and name alignment. A **challenge** (replacement) requires a confident challenger that disagrees by >100 ft, wins on both evidence axes, *and* an incumbent that is geometrically indefensible (verification < 0.1). A **refinement** adopts a challenger that *agrees* (<100 ft) and beats the incumbent's verification by ≥ 0.1 — measured on truth, an agreeing snap pose beats a 25–50 ft RANSAC fit 93% of the time (median 33 → 12 ft) and a 50–100 ft fit 85% of the time, because RANSAC's mid-tier error is street-label placement noise while the snap pose is chamfer-locked to the street grid.

**Production surface.** `mapsnap snap DIR` runs candidates → select → materialize at the frozen gates and writes `pN.georef-osm.json` sidecars. `mapsnap fit` invokes it after `georef` and builds the IIIF from the osm-first hybrid glob (`*.georef-osm.json,*.georef.json` — first glob wins per page), which carries the channel into `run-oim`, `run-loc`, and `rerun` unchanged. Opt out with `--no-snap`. First run per volume costs UNet inference + matching (~10–30 min); candidates are cached in `artifacts/osm_snap/candidates.jsonl` and reruns take seconds.

## Illustrations

(Image files are staged locally in `pr-images/` — drag them into the placeholders below. See also the volume-viewer DC progression in the comments.)

**A clean rescue lock.** DC p130, unplaceable by street-name OCR, snapped at 4 ft: the page's road-UNet corridors (red) fuse with the OSM grid (green) into yellow across the entire footprint.

01-rescue-lock-dc-p130-4ft

**Why gating matters — a confident grid alias.** Detroit p85 at 368 ft: the page's uniform corridors lock onto *alternating* OSM streets. The interior aliases perfectly; only the page edges (the top corridor with no green partner) give it away.

02-grid-alias-detroit-p85-368ft

**Refinement.** DC p201: the RANSAC incumbent (left) sits 87 ft off — every corridor slightly misregistered; the adopted agreeing challenger (right) locks to 12 ft. This one move, repeated 274 times, is the +8.7-point step.

03-refinement-dc-p201-87ft-to-12ft

**Split sheets are inset composites.** Champaign p4's four panels' true placements: scattered kilometres apart, one rotated — the paper layout has no relation to ground geometry, which is why the sheet-agreement gate acts as a block rather than a geometric check (#154).

04-inset-composite-champaign-p4

**Channel architecture.**

```mermaid
flowchart LR
    G["mapsnap georef
(RANSAC on street OCR)"] --> S
    subgraph S["mapsnap snap"]
        C["candidates
P(road) vs OSM raster:
prior ladders → NCC → chamfer
+ evaluate_pose(incumbent)"] --> SEL
        subgraph SEL["select --mode arbitrate"]
            R["rescue
(unplaced pages,
union of 2 committees)"]
            A["arbitrate
(incumbent ver < 0.1,
disagree > 100 ft)"]
            F["refine
(agree < 100 ft,
ver margin ≥ 0.1)"]
        end
        SEL --> M["materialize
pN.georef-osm.json"]
    end
    M --> I["mapsnap iiif
osm-first hybrid glob"]
    G --> I
    I --> CMP["mapsnap compare / score"]
```

## Glossary

- **P(road)** — per-pixel road probability from the road-UNet (#125), cached as `artifacts/edge_join/roadprob/*.png`.
- **Snap / snap pose** — a placement obtained by matching P(road) against rasterized OSM geometry.
- **Incumbent / challenger** — the existing RANSAC fit for a page / the top snap candidate for the same page.
- **Rescue** — placing a page that has no `pN.georef.json` (nofit/misscale/1gcp/outlier variants, or nothing at all).
- **Arbitration** — replacing an incumbent that OSM contradicts with a strongly disagreeing challenger.
- **Refinement** — adopting an agreeing challenger as a more precise estimate of the same lock.
- **Verification** — the matcher's truth-free pose quality: chamfer inlier fraction + fine NCC − mean chamfer/30 m.
- **select_score** — verification + name-alignment + containment + prior-agreement bonuses; the ranking/gating score.
- **Distinct margin** — select_score lead of rank-1 over the best candidate that is a genuinely different lock (>100 m or >10° away).
- **NCC** — normalized cross-correlation (Padfield masked FFT form) used for the translation search.
- **Chamfer** — the page's road-skeleton points projected through a pose and read against the OSM centerline distance transform: the *mean* (clamped at 30 m) measures overall fit, and the **inlier fraction** is the share of skeleton points landing within 5 m of an OSM centerline — how much of the page's road network locks onto the grid.
- **Keymap** — the volume's index sheet; its georeferenced page-number detections and segmented block regions give each page a coarse location and footprint.
- **Sheet agreement** — the split-panel gate. A panel jpg is a bounding-box crop of the base sheet, so one panel's pose implies a placement of the whole scanned sheet; the gate demands siblings imply the same one. Because real split sheets are inset composites (see lessons learned), this in practice blocks sibling-of-fitted panels from rescue rather than checking geometry.
- **eff / effective GCPs** — effective ground-control-point count of a fit; ≥3 marks a fit reliable enough to anchor sheet agreement.
- **IoU-over-min** — intersection area over the smaller footprint's area; the overlap measure in the volume energy.
- **Net score** — the `mapsnap score` metric: land-weighted share of truth pages ≤25 ft minus share ≥200 ft (#148).
- **GEOREF_VARIANTS** — sidecar suffixes (`georef`, `-nofit`, `-misscale`, `-1gcp`, `-outlier`, `-neighbor`, `-osm`) recording each page's fit state.
- **OIM** — OldInsuranceMaps.net, the source of truth annotations (`main.iiif.json`) and manual split regions.

## Lessons learned / dead ends

- **Truth data had two silent holes**, both now fixed and guarded: (1) truth keys differing from jpg stems only in letter case left 111 Chicago pages invisible to the experiment harness; (2) four volumes had truth split items but no `oim/*.panels.json`, so *no placement of those pages could ever be scored* — KC's real baseline was ~4 points higher than reported. `mapsnap compare` (and through it `score`) now warns loudly on the second condition; `mapsnap oim-split-truth` inputs for KC/Detroit/NO-1896 were fetched and built.
- **Score alone cannot gate small pages.** Split panels produced confident wrong locks (847 ft at select 2.28) interleaved with true locks. Contiguity with the fitted sibling was not enough — an alias slid *along the cut line* still touches it. The rigid-sheet agreement constraint (exact, geometric) was the fix, and it later blocked four would-be KC disasters including a 753 ft lock at select 2.58.
- **Fitted split panels are the least reliable fits in a volume** (eff ≤ 2 panels range to 6640 ft off). They are excluded from the overlap-evidence context and from sheet-agreement anchoring; agreeing with a bad anchor rejects true candidates.
- **The modern-OSM trap**: in Chicago's rebuilt 1950 core, a 400 ft-wrong pose can match *today's* OSM better than the truth does, with name evidence tied. This produced the only bad arbitration challenges and motivated the incumbent-defensibility gate (challenge only fits with verification < 0.1). Across all 11 truth-graded challenges the gate separates perfectly (correct overturns ≤ 0.05, wrong ones ≥ 0.12).
- **A stronger optimizer made scores worse.** Adding ICM restarts found the volume energy's true optimum — which abstains on pages the accept-biased two-start version correctly keeps, because the adjacency term over-penalizes true placements where key-map centroids are unreliable (LA's lettered sheets). The two-start bias is kept deliberately, with a comment.
- **Calibrate against the volume's own geometry, not constants**: the keymap search radius (p90 of fitted residuals), the overlap thresholds (p90 of fitted seam overlaps, one value per number pair — exhaustive member combos flood the distribution with near-zeros), and the containment gate (0.35 borrowed from edge-join killed true 10 ft locks; schematic regions run small).
- **Directed priors only for flip detection**: the mod-180 rotation rung emits both flips, so including it in the prior-agreement feature pinned the residual at zero and masked 180° flips.
- **Small pages need a proportional NCC floor**: the fixed 30,000 m² minimum overlap exceeded some panels' entire ground area, silently masking every shift.
- **The rigid-sheet premise is false for real splits** (post-merge-review finding): split pages are inset composites whose paper layout is unrelated to ground geometry — every fitted sibling pair measured disagrees on its implied sheet by 233–2143 m. The sheet-agreement gate therefore protects by blanket-blocking, not by discrimination; its net effect is validated, its label was wrong.
- Earlier related dead ends that shaped this design: edge-to-edge page joins alone (NO-GO on LA — seam quality, not connectivity, was the limiter), and keymap-coupled adjacency for the pose graph (connectivity worked, disasters neutralized the gains). The discrete-candidates-plus-gated-selection shape here is what those experiments pointed to.

## Debugging

Each page's snap decision (priors, top-3 candidate evidence, incumbent evaluation, outcome) is appended to the same `pN.txt` log the RANSAC georeferencer writes, marker-delimited, so the debugger's log view covers both channels. Per volume: `snap-base` / `snap-union2` / `snap-arb` / `snap-refine` `.iiif.json` + `.txt` render the journey stages in the debugger (`&files=` interface). `artifacts/osm_snap/candidates.jsonl` holds every candidate with full feature vectors and truth RMSE where available; `selection_*.jsonl` the per-mode decisions; `vis/` red/green contact sheets of top-2 candidates per page. Intermediate experiment tags (`snap-v0`, `snap-v1`, `snap-v1b`, `snap-union`) and the 1-GCP ablation tags (`exp-with1gcp`, `exp-no1gcp`) also remain on disk.

## Next steps

- **Re-sweep the refinement margin** (#153) with the larger graded-challenge population now available; the current 0.1 was a small-sample dev elbow, and some agreeing-but-unadopted challengers in 25–100 ft likely remain capturable.
- **Recall residual**: Hudson's OSM-divergent waterfront (27.7% of its land has no matchable modern geometry), no-keymap pages (Nashville 12.7% — a keymap-read fix, e.g. GR's leading-'7' truncation), and Detroit's translational grid aliasing (v2 pose-graph fusion of snap candidates as absolute-pose hypotheses is designed but unbuilt).
- **Remaining disaster tail** (2.9%): all pre-existing RANSAC placements; per-volume `oim-split-truth` for volumes beyond the three fixed here would also let more split placements be credited.
- **Per-inset panel gating** (#154): replace the sheet-agreement fiction with per-inset treatment (each panel gated independently on keymap containment/solo score, with the blanket block on sibling-of-fitted panels kept until something beats it empirically); auto-detecting the rare contiguous-split sheet via fitted-sibling agreement is the cheap first step.
- **Runtime** (#155): the matching sweep (not UNet inference) dominates the ~10–30 min/volume. Cheap algorithmic wins first (theta-ladder pruning under confident priors, shared OSM rasters between incumbent evaluation and matching, an STRtree cull for giant centerline files), then page-level multiprocessing, then optionally batched GPU FFT correlation.
- **Gate calibration debt**, disclosed: the incumbent-defensibility threshold and the panel solo gate were calibrated on all available graded events (11 and ~14 respectively — too few for clean dev/holdout splits). More truth volumes would firm these up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,10 +325,11 @@ The four arguments here are:
 - `mapsnap ocr`: runs OCR over the downscaled images, saving candidate detections to `boxes.json` and `streets.json` files.
 - `mapsnap fit`: another pipeline that runs:
   - `mapsnap georef`: georeferences images based on street detections, writing out `georef.json` files where it can find a good fit.
+  - `mapsnap snap`: the geometry-first channel — matches each page's road-UNet mask directly against OSM street geometry to rescue pages the street-name georeferencer couldn't place, replace fits that OSM actively contradicts, and refine mid-tier fits to street-grid precision. Writes `georef-osm.json` sidecars that take priority in the next step. Skip with `--no-snap`.
   - `mapsnap iiif`: produces a IIIF Georeference Extension with clipping masks. You can find examples of these in the `gallery` directory. View them on Allmaps.
   - `mapsnap compare`: compares the generated IIIF file with the human-generated one from OIM, producing a report on the accuracy of the fit.
 
-`mapsnap ocr` is typically the slowest step. The steps under `mapsnap fit` run relatively quickly. You can run `mapsnap fit` yourself to experiment.
+`mapsnap ocr` is typically the slowest step. The first `mapsnap snap` run on a volume also takes a while (road-UNet inference plus grid matching, roughly 10–30 minutes); its candidates are cached and later runs take seconds. The other steps under `mapsnap fit` run relatively quickly. You can run `mapsnap fit` yourself to experiment.
 
 The end result is a IIIF file pointing at Library of Congress imagery that you can view in Allmaps, along with a text file comparing Mapsnap's results against OIM's.
 
--- a/mapsnap/cli.py
+++ b/mapsnap/cli.py
@@ -43,6 +43,10 @@ SUBCOMMANDS: dict[str, tuple[str, str]] = {
         "mapsnap.page_adjacency",
         "Detect printed adjacent-sheet numbers and build a volume adjacency graph",
     ),
+    "snap": (
+        "mapsnap.snap",
+        "Geometry-first OSM snap: rescue unplaced pages, arbitrate and refine fits",
+    ),
     "iiif": (
         "mapsnap.make_iiif_georef",
         "Combine georeferences into a IIIF AnnotationPage",
--- a/mapsnap/compare_iiif_georef.py
+++ b/mapsnap/compare_iiif_georef.py
@@ -713,6 +713,20 @@ def compare_pages(
         source = truth_items[0]["target"]["source"]
         source_dims = (float(source["width"]), float(source["height"]))
         truth_polygons = load_split_polygons(oim_dir / f"{page_key}.panels.json")
+        if not truth_polygons:
+            # Without OIM's panel polygons NO placement of this page can ever
+            # be matched to its truth splits — every one scores as unplaced,
+            # silently understating the volume (KC sat ~4 net points low this
+            # way). Loud warning rather than assertion: the fix is a data run
+            # (`mapsnap oim-split-truth`), and comparison of the other pages
+            # is still meaningful in the meantime.
+            print(
+                f"Warning: {page_key} has {len(truth_splits)} truth split(s) but "
+                f"{oim_dir / f'{page_key}.panels.json'} is missing; its placements "
+                "will score as unplaced. Run `mapsnap oim-split-truth` (see its "
+                "docstring for the required raw/ and oim/ images).",
+                file=sys.stderr,
+            )
         gen_polygons = load_split_polygons(
             gen_dir / f"{page_key}.panels.json", source_dims
         )
--- a/mapsnap/compare_iiif_georef_test.py
+++ b/mapsnap/compare_iiif_georef_test.py
@@ -342,3 +342,46 @@ def test_compare_pages_ignores_skeletons_with_full_color_counterparts(tmp_path):
     # it must not count as a missing page.
     assert [r["page_key"] for r in rows] == ["p153"]
     assert missing == []
+
+
+def test_compare_pages_warns_on_truth_splits_without_oim_panels(tmp_path, capsys):
+    from mapsnap.compare_iiif_georef import compare_pages
+
+    def item(page_id: str, label: str) -> dict:
+        return {
+            "label": label,
+            "target": {
+                "source": {
+                    "id": f"https://loc.gov/x/g123-{page_id}/info.json",
+                    "width": 1000,
+                    "height": 1000,
+                },
+            },
+            "body": {
+                "transformation": {"type": "polynomial", "options": {"order": 1}},
+                "features": [
+                    {
+                        "properties": {"resourceCoords": [x, y]},
+                        "geometry": {
+                            "coordinates": [-74.0 + x * 1e-6, 40.0 + y * 1e-6]
+                        },
+                    }
+                    for x, y in [(0, 0), (900, 0), (0, 900)]
+                ],
+            },
+        }
+
+    # Truth split items but no oim/p52.panels.json: the generated placement can
+    # never be matched — that silent gap once understated KC by ~4 net points,
+    # so compare (and through it `mapsnap score`) must say so out loud.
+    truth = {"items": [item("0052", "x p52 [1]"), item("0052", "x p52 [2]")]}
+    generated = {"items": [item("0052", "x p52")]}
+    truth_path = tmp_path / "main.iiif.json"
+    gen_path = tmp_path / "gen.iiif.json"
+    truth_path.write_text(json.dumps(truth))
+    gen_path.write_text(json.dumps(generated))
+    rows, missing = compare_pages(truth_path, gen_path)
+    assert rows == []
+    assert len(missing) == 2
+    err = capsys.readouterr().err
+    assert "p52" in err and "oim-split-truth" in err
--- a/mapsnap/edge_join.py
+++ b/mapsnap/edge_join.py
@@ -71,6 +71,60 @@ def compose(a: np.ndarray, b: np.ndarray) -> np.ndarray:
     return a @ np.vstack([b, [0.0, 0.0, 1.0]])
 
 
+@dataclass
+class FrameSpec:
+    """A pair-local raster frame: equirectangular metres, north-up rows.
+
+    col = (x - x_min) / res, row = (y_max - y) / res, where (x, y) are metres
+    east/north of `origin`. Page images map into the raster without reflection.
+    """
+
+    origin: tuple[float, float]  # lon, lat
+    x_min: float
+    y_max: float
+    res_m: float
+    shape: tuple[int, int]  # rows, cols
+
+    def metre_scales(self) -> tuple[float, float]:
+        kx = 111_320.0 * math.cos(math.radians(self.origin[1]))
+        return kx, 110_540.0
+
+    def lonlat_to_raster(self, lon: float, lat: float) -> tuple[float, float]:
+        kx, ky = self.metre_scales()
+        x = (lon - self.origin[0]) * kx
+        y = (lat - self.origin[1]) * ky
+        return (x - self.x_min) / self.res_m, (self.y_max - y) / self.res_m
+
+    def page_to_raster_affine(self, affine_local: np.ndarray) -> np.ndarray:
+        """2x3 mapping page px -> raster px given a page px -> lon/lat affine."""
+        kx, ky = self.metre_scales()
+        a = affine_local
+        metres = np.array(
+            [
+                [a[0, 0] * kx, a[0, 1] * kx, (a[0, 2] - self.origin[0]) * kx],
+                [a[1, 0] * ky, a[1, 1] * ky, (a[1, 2] - self.origin[1]) * ky],
+            ]
+        )
+        to_frame = np.array(
+            [
+                [1 / self.res_m, 0, -self.x_min / self.res_m],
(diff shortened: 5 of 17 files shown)


danvk:
This initially looked like a bit of a wash on everything but DC, but adding geometry-first georeferencing on existing RANSAC fits ("refine") made it a home run. Geometry matching seem very good at turning a rough location into a precise location.

Here's DC on snap-base:

washington-snap-base-fs8

Here's DC on snap-arb:

washington-snap-arb-fs8

and finally snap-refine:

washington-snap-refine-fs8

Issue #154, opened 2026-07-24
osm-snap: per-inset panel gating (split sheets are inset composites, not contiguous maps)


danvk:
The split-panel "sheet-agreement" gate (`panel_allowed_candidates` / `sheets_agree` in `osm_snap_experiment.py`, `SHEET_AGREE_TOL_M = 60`) was designed on the premise that a split page is one contiguous map cut into pieces, so all panels imply the same full-sheet placement. **Measured against every fitted sibling pair in the 12 truth volumes, that premise is false**: real Sanborn split pages are inset composites — separate neighborhood maps pasted onto one sheet of paper — whose implied sheets disagree by 233–2143 m (Champaign p20: 738 m, p21: 2.1 km, DC p260: 1.8 km).

Since correct placements also fail the 60 m agreement test, the gate in practice **blanket-blocks rescue of any panel with a reliably fitted sibling (eff ≥ 3)**. That conservative behavior is what the 12-volume scores validated — it's what stopped KC's four would-be panel disasters, including p555__2 at 753 ft with select 2.58, a score no threshold would catch — but it also forfeits blocked-but-correct panels: KC p566__2 (8.3 ft rank-1), Champaign p13__2 (6.7 ft), DC p213__2 (30 ft).

**Proposed redesign:**
1. **Auto-detect the sheet class**: when ≥2 siblings are reliably fitted, compare their implied sheets. Agreement ≤ ~60 m → genuinely contiguous split → apply the rigid-sheet gate as designed (it is exact for that class). Disagreement → inset composite.
2. **Per-inset gating for composites**: gate each panel independently — key-map region containment (the sheet's region covers all its insets), the existing solo score bar, and possibly a name-alignment requirement — instead of the blanket block.
3. **Validate on the panel truth set** (KC/GR/DC/LA/Champaign/Nashville now have `oim/*.panels.json`): the danger case p555__2 (753 ft @ select 2.58) must stay blocked; the blocked-correct panels above should be recovered.

Truth-annotated panel candidates are already cached in `artifacts/osm_snap/candidates.jsonl`, so this is a selection-layer experiment — no re-matching needed.

danvk:
Measured the proposed redesign against the cached panel candidates (truth RMSE attached, no re-matching). **It does not survive contact with the data — I have not implemented it.** Three reasons, then what I'd suggest instead.

## 1. The sheet-class auto-detection has no data to run on

[...]

PR #160, opened 2026-07-24
Score truth split panels over their own regions (bad-split sensitivity)

danvk:
Makes the score metric sensitive to bad splits, in both directions. Previously these two outcomes scored **identically**:

- **Scenario A**: the splitter fails (or a split page is georeferenced whole), and the page lands on one inset of a composite sheet. The other insets' imagery is *displayed kilometres from where it belongs* — but their truth panels just scored "unplaced."
- **Scenario B**: the splitter works, one panel is placed, the others are honestly abstained. Also "unplaced."

Scenario B is objectively better; now the metric agrees.

## What changed

Truth split panels are no longer paired one-to-one with generated annotations (with an unsplit placement standing in for the *largest* truth panel). Instead, **every truth panel is graded over its own region**: the RMSE grid is sampled inside the panel's `oim/pN.panels.json` polygon, against whichever generated annotation **covers** that region (one-to-many; coverage below `TRUTH_REGION_COVERAGE_MIN = 0.5` counts as unplaced). `match_split_pairs` is superseded and removed; three scenario tests pin the semantics.

Three concrete effects:
1. **Wrong-content display becomes a disaster** (Scenario A). A whole-page or wrong-inset placement is graded on every truth region it covers; regions displayed under the wrong transform grade at their true (huge) error instead of vanishing into "unplaced."
2. **Correct coverage earns full credit.** A whole-page placement that correctly displays several truth panels is credited for each of them (the old rule credited only the largest).
3. **Matched panels are measured where their content actually is.** The old metric sampled the grid over the *generated* annotation's split-canvas rectangle (or the whole canvas), which extends past the truth panel into masked-out or extrapolated area — and affine-vs-affine disagreement grows with distance from the GCP centroid, inflating RMSE. The new grid stays inside the truth polygon.

## Why a score could go UP: Champaign, in detail

Champaign moves **92.7 → 97.8** (+5.1), which looks paradoxical for a stricter metric until you trace it. Two pages tell the whole story:

**p21__2 — effect 3, worth the entire +5.1.** Both of p21's panels were split and RANSAC-fitted all along. Under the old metric the p21__2 pair was graded over *our* panel's split-canvas rectangle, whose cut lines differ from OIM's — so the 7×7 grid included canvas outside the truth panel, where the two affines are extrapolated apart. Result: 27.4 ft, a hair **over** the 25 ft good bar, worth zero. Graded inside the truth panel's own polygon — where the imagery it's responsible for actually lies — the same placement measures **18.2 ft**: good. p21__2 happens to be the volume's largest panel (5.3% of Champaign's land), and one threshold crossing on a big panel in a 33-page volume is +5.3 good share. **The placement never changed; only where it was measured.** The old score under-credited it, which is the answer to "did I misunderstand the old score": for split pairs, the old grading region was the *generated* annotation's display rectangle, not the truth panel — a subtle asymmetry that could penalize correct fits for canvas they weren't responsible for.

**p3 — Scenario A itself, newly visible.** p3 is a three-inset composite sheet that our pipeline georeferenced **whole**, landing (well) on the middle inset: truth p3__2 grades at 15 ft under both metrics. But the whole-page transform also *displays* insets 1 and 3, kilometres from their true locations. Old metric: the whole-canvas stand-in paired only with the largest panel; p3__1 and p3__3 scored unplaced — indistinguishable from honest abstention. New metric: their regions are covered, so they grade — at **7,411 ft and 2,159 ft**, disasters. They're small panels (0.1% of land each), so the penalty is −0.2; the p21__2 correction dominates and Champaign nets +5.1.

The same decomposition explains every mover: Grand Rapids +2.0 (effect 2/3 credits on split sheets), LA **−1.4** (effect 1: five panels displayed under the wrong inset's transform — p1442__1 at 10,239 ft, p1470__2 at 12,110 ft, and the p1499K/L family at 900–8,900 ft — previously invisible, now costed).

## Per-volume movement (final `snap-refine` configuration)

| Volume | old | new | Δ |
|---|---|---|---|
| Champaign | 92.7 | **97.8** | +5.1 |
| Grand Rapids | 64.0 | **66.0** | +2.0 |
| Los Angeles | 76.1 | **74.7** | −1.4 |
| Kansas City | 68.3 | 68.2 | −0.1 |
| Washington DC | 76.1 | 76.0 | −0.1 |
| Brooklyn, Chicago, Detroit, Hudson, Nashville, NO-1896, NO-1951 | — | — | 0.0 |
| **AGGREGATE** | **69.8** | **70.0** | **+0.2** |

The baseline shifts too (55.5 → 55.7), so the snap channel's measured contribution is unchanged. This is a **re-baseline**: numbers in older discussions/PRs use the previous metric.

## Robustness: non-convex panels

Review of Champaign's p21 surfaced a sampling subtlety: its second panel is an **L-shape** filling 57% of its bounding box, so a 7×7 bbox lattice keeps only ~24 interior points and understates the converged RMSE by ~6% (18.2 ft at n=7 vs 19.3 ft at 2,121 points — still safely under the bar, so Champaign's number stands). The grid now **densifies until non-convex panels get at least the 49 interior points a rectangle gets**. That honesty pass moved Grand Rapids 67.7 → 66.0: a third of its apparent gain was sparse-sampling luck on panels near the threshold. A test pins the L-shape density.

## Follow-ups this enables

- LA's newly visible wrong-content panels are an actionable target class (they overlap heavily with #154's per-inset panel work, which now has score incentive attached).
- The metric now polices the splitter itself: regressions that stop splitting composite sheets cost points instead of hiding.
- The experiment harness's `attach_missing_truth` still uses the largest-split approximation for per-candidate diagnostics (noted in its docstring); the scorer is the arbiter.

### DIFF
--- a/mapsnap/compare_iiif_georef.py
+++ b/mapsnap/compare_iiif_georef.py
@@ -27,6 +27,7 @@ from collections import defaultdict
 from pathlib import Path
 
 import numpy as np
+from shapely.geometry import Point as ShapelyPoint
 from shapely.geometry import Polygon as ShapelyPolygon
 from shapely.geometry import box
 
@@ -37,6 +38,9 @@ EARTH_RADIUS_FT = 20_925_524.0
 MIN_SPLIT_IOU = (
     0.1  # minimum panel overlap to treat a truth and generated split as the same region
 )
+# A truth panel is graded only when this fraction of its region is actually
+# displayed by some generated annotation; below it, the panel is unplaced.
+TRUTH_REGION_COVERAGE_MIN = 0.5
 
 
 def extract_metadata_int(item: dict, label: str) -> int | None:
@@ -248,6 +252,26 @@ def sample_grid(width: float, height: float, n: int = 7) -> list[tuple[float, fl
     return [(f_x * width, f_y * height) for f_y in fracs for f_x in fracs]
 
 
+def truth_region_grid(polygon: ShapelyPolygon, n: int = 7) -> list[tuple[float, float]]:
+    """Grid sample points inside a truth panel's polygon (canvas pixels).
+
+    Samples an n×n lattice over the polygon's bounds (with the same
+    corner-avoiding fractions as sample_grid) and keeps the interior points,
+    so a split panel is graded over ITS OWN region rather than the whole
+    canvas. Falls back to the full lattice for degenerate slivers that
+    contain fewer than five lattice points.
+    """
+    min_x, min_y, max_x, max_y = polygon.bounds
+    fracs = [(k + 1) / (n + 1) for k in range(n)]
+    lattice = [
+        (min_x + f_x * (max_x - min_x), min_y + f_y * (max_y - min_y))
+        for f_y in fracs
+        for f_x in fracs
+    ]
+    inside = [p for p in lattice if polygon.contains(ShapelyPoint(p))]
+    return inside if len(inside) >= 5 else lattice
+
+
 def truth_distortion(A: np.ndarray) -> tuple[float, float]:
     """Compute skew (°) and anisotropy from the truth affine linear part.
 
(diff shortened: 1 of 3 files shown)


danvk:
Here's p21__1, the L-shaped split:
p21__1

PR #170, opened 2026-07-28
Adjacency-truth labeler in the debugger app

danvk:
Adds a truth-labeling mode for printed adjacent-sheet numbers, reusing the key-map labeler's components. Pick a volume, pick a page, click the centers of the printed neighbor numbers along its edges, and type each number — labels save automatically to the volume's `adjacency-truth.json` (a sibling of the pipeline's `adjacency.json`, keyed by page stem, coordinates in the volume-root image frame).

- **Server**: typed crosswalk API (`/api/adjacency-volumes|pages|labels`) with volume/page-stem validation, natural page sort, and read-modify-write into one truth file per volume. Page images are served through the existing static `data/` route.
- **App**: volume selector with labeling progress, pages list, click-to-place labels with size-slider preview boxes, n/p/j/k page paging, a "Next page" button reachable by tab+enter that focuses the next page's first text box, and per-page blue/orange counts (labels with/without text) shared with the key-map labeler.
- **Saves**: debounced, with an immediate flush before any page/volume switch so edits made inside the debounce window are never dropped.

First use: complete truth for Hudson Co. (92 pages, 463 printed references). Grading against it showed the mutual-edge detector at 100% pair precision / 74% recall of mutually-printed pairs, and validated the OIM-footprint proxy in `score_adjacency.py` (its verdicts matched hand truth on all 165 detector edges; its recall denominator overcounts by ~13% from corner contacts sheets don't print).

### DIFF
--- a/app/adjacency.html
+++ b/app/adjacency.html
@@ -1,0 +1,12 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Adjacency Labeler</title>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script type="module" src="/src/adjacency-main.tsx"></script>
+  </body>
+</html>
--- a/app/server/adjacencyRoutes.ts
+++ b/app/server/adjacencyRoutes.ts
@@ -1,0 +1,80 @@
+/**
+ * Typed JSON API for the adjacency-truth labeler (`/api/adjacency-*`).
+ *
+ * Page images are not served here: the labeler loads them through the static
+ * `data/` route (the Vite serveDataDir plugin in development, express.static
+ * in production), which both resolve `data/<volume>/<stem>.jpg` relative to
+ * the app base.
+ */
+
+import { HTTPError, type TypedRouter } from 'crosswalk';
+import type { API } from './api.ts';
+import {
+  findVolumes,
+  isSafePage,
+  isSafeVolume,
+  readTruth,
+  volumePages,
+  writePageTruth,
+} from './adjacencyTruth.ts';
+import type { ImageInfo } from '../src/keymap/types.ts';
+
+/** Register the adjacency-truth endpoints on the shared crosswalk router. */
+export function registerAdjacencyTruthApi(
+  router: TypedRouter<API>,
+  dataDir: string,
+): void {
+  // Volumes that have labelable pages, with their labelling progress.
+  router.get('/api/adjacency-volumes', async () => {
+    const volumes = [];
+    for (const name of await findVolumes(dataDir)) {
+      const pages = await volumePages(dataDir, name);
+      if (!pages.length) continue;
+      const truth = await readTruth(dataDir, name);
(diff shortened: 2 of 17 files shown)


danvk:
Details on the OIM vs. real truth eval below. re: "p86 and p92 each have two labels bearing their own page number," those are referring to other splits of the same page. I haven't checked the suspicious truth data it flagged yet.

Prompt (2026-07-27T22:18:50.532Z):

> While that runs… I've generated complete adjacency truth data for Hudson Co. There's existing code to score adjacency detections in @mapsnap/score_adjacency.py . This uses truth data to verify adjacency. Now that we have real truth data, though, I'd like you to compare the detections to the truth data and see how accurate the OIM-derived adjacency was all along.

Response (claude-fable-5):

The truth data settles it — the OIM proxy was an excellent verifier all along. Full picture:

## Hand truth (92 pages, 463 labels)

447 directed printed references → **230 unordered pairs**: 215 printed on both sheets, 15 one-sided. Two artifacts to flag: **p86 and p92 each have two labels bearing their own page number** — (726,1758)/(173,1465) on p86, (888,1873)/(1161,1388) on p92. Probably the sheet's own number printed at panel corners got clicked; worth a 10-second look.

## How accurate was the OIM-footprint proxy?

| | vs hand truth |
|---|---|
| Proxy pairs (30m gap) that are actually printed | 228/262 (**87%**) |
| Printed pairs the proxy covers | 228/229 (**99%**) |
| Verdicts on the detector's 165 edges | **165/165 agree** — 0 overcredits, 0 false penalties |

- The docstring's two caveats are now quantified. The **corner-contact recall drag is real**: all 34 unprinted proxy pairs are short contacts (median ~27m ≈ one block; max p1–p4 at 98m) that sheets print no reference for. Old reported recall 165/262 = 63%; **true printed-pair recall is 165/230 = 72%** — the proxy understated recall by ~9 points.
- The **"precision reads high" worry never materialized** on Hudson: every proxy-credited edge is genuinely printed. Precision numbers from `score_adjacency` were exact, not optimistic.
- The single proxy miss: **p44 ↔ p46**, printed (one-sided) but 185m apart in OIM truth — a genuine across-a-gap reference no gap threshold ≤60m catches. Gap choice barely matters otherwise (10m→60m: 258→267 pairs).

## Detector graded against real truth for the first time

- **Directed claims**: 372/428 precision (87%), 372/447 recall (83%) of printed refs.
- **Mutual edges**: 165 edges, **100% correct** against either-sheet-printed truth (96% against both-printed); recall 159/215 = **74%** of mutually-printed pairs.
- The 6 edges where truth says only one sheet prints the reference — p3↔p6, p22↔p24, p27↔p28, p36↔p37, p43↔p44, p59↔p60 — are cases where the detector read a number on *both* sheets. Given claim precision, these are more likely **truth labels the labeler missed** (faint prints) than lucky misreads; the labeler's page view will confirm quickly.
- Structural note for the LA one-sided-claims question in memory: 15/230 Hudson pairs (7%) are printed one-sided only, so a mutual-edge design forfeits them by construction — reciprocity's cost is real but modest here.

Net: on Hudson, `score_adjacency`'s proxy needed no apology on precision and only a denominator correction on recall. The keymap rerun is still going in the background; I'll report when it lands.

PR #171, opened 2026-07-28
Full-string page keys: lettered vocabulary from disk through the KeymapLocator

danvk:
Page vocabularies, key-map detection texts, and `KeymapLocator` keys become full page-key strings (digits + optional letter suffix, uppercase) instead of bare ints, so letter-suffixed pages are first-class throughout the keymap channel.

## Why

- The int-keyed locator conflates letter families: LA's 18-page `1499`/`1499A`–`R` family and asheville's `3`/`33`/`35` collisions (21 pages) union into single junk locations today.
- The int-only `--pages` spec misreads lettered sheets (known new-volume blocker).
- Chicago's disk keys are all direction-suffixed (`51N`, `100W`) while its sheet prints bare digits — no detection can currently carry the true page key at all.

## What changed

- The valid page set is now **derived from each image's own volume** (its `p*.jpg` stems, letters included) — per image, so one invocation spanning volumes can never mix page sets. `--pages` becomes an override. `parse_page_spec` takes lettered tokens alongside numeric ranges.
- `snap_to_pages` gets principled: exact match → unique same-stem key (the letter comes from the vocabulary) → unique key within edit distance 1, plus a unique leading-digit completion for truncated reads. **Arbitrary tie-breaks are gone**: a read one edit from several keys is kept raw rather than assigned a valid-but-wrong key — the class that poisoned KC p470's keymap location (961m off) in the last pipeline rerun.
- `KeymapLocator` stores string keys; lookups accept `int | str`. A string resolves exact-first then stem-family; an int resolves to the whole family — exactly the old behavior, so int-keyed consumers (`osm_snap`, `edge_join`, `align_page_region`) are semantically unchanged. `ocr`/`georef` look up by the page's disk key. `load_detections` keeps lettered texts (was digits-only).

## Per-volume impact (full detector rerun vs current sidecars, same models)

Metric: **key-correct** — the detection text equals the disk page key at the truth label's location (printed `51` on chicago denotes disk key `51N`; nashville's printed `54A` denotes disk key `54`).

| volume | before | after |
|---|---|---|
| chicago | 0/112 | **99/112** |
| queens vol3 | 104/113 | 106/113 |
| brooklyn 1906 | 78/84 | 79/84 |
| new_orleans 1896 | 87/90 | 88/90 |
| champaign, GR, hudson, KC, DC | — | −1 each |
| san_francisco | 109/114 | 106/114 |
| 12 other sheets | — | unchanged |
| **total** | **1268/1609 (78.8%)** | **1363/1609 (84.7%)** |

The −1s and SF's −3 are the removed tie-breaks giving back their occasional lucky wins (plus some CRNN run-to-run variance): e.g. GR's raw `81` used to snap arbitrarily into `810`–`819` (right once, wrong five times). The safety trade is measured directly: **valid-but-wrong keys — the ones that poison locator locations — drop 119 → 96**, and total matched-but-wrong texts drop 227 → 131. A kept-raw read is inert downstream; a valid-but-wrong key corrupts a page's location.

## Verification

745+ tests (new coverage for lettered specs, snap rules, locator family fallback, vocabulary derivation), pyright clean. Existing sidecars with bare-digit texts keep working via the stem-family fallback; lettered keys appear as volumes re-run `mapsnap keymap`.

### DIFF
--- a/mapsnap/detect_text.py
+++ b/mapsnap/detect_text.py
@@ -17,7 +17,7 @@ from PIL import Image
 from tqdm import tqdm
 
 from mapsnap.ctc_vocab_decode import HINT_STRINGS, generate_vocab_strings
-from mapsnap.keymap.locate import KeymapLocator, page_number, resolve_keymaps
+from mapsnap.keymap.locate import KeymapLocator, page_key, resolve_keymaps
 from mapsnap.streets import build_block_index, polygon_side_lengths
 from mapsnap.utils import default_centerlines, image_stem
 
@@ -735,7 +735,7 @@ def page_vocabs(
     if locator is None:
         return vocab_strings, None
     restricted = locator.restricted_features(
-        page_number(image_stem(image_path)), geojson_features
+        page_key(image_stem(image_path)), geojson_features
     )
     if not restricted:
         # Unplaced (None) or placed with no nearby features ([]): rectangle is the tightest.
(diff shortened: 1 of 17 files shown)


danvk:
`--pages` could probably be removed at this point

Issue #168, opened 2026-07-27
Fit using streets, not GCPs


danvk:
The advantage of GCPs is that they're at specific points, whereas streets can curve. But when we have a coarse location, streets often stay straight within the keymap radius. Depending on connectivity, fitting on the streets alone may work significantly better than requiring GCPs.

New Orleans 1896 p181 is the motivating example:

<img width="803" height="953" alt="Image" src="https://github.com/user-attachments/assets/bec3b57c-bd11-4ca3-83d8-22ee150af1d5" />

This produces a rotated fit because LAFAYETTE is the only street that produces GCPs, and it's slightly tilted by #3.

<img width="708" height="863" alt="Image" src="https://github.com/user-attachments/assets/49bc473c-e3b0-48a6-bcb4-4d67ac33facc" />

But JOHNSON, GALVEZ, TONTI and LAFAYETTE are enough to set the model all on their own. They don't need to intersect.

Filtering outliers may be a bit more of a challenge in this model since the existence of a GCP implies some kind of coherence to the detections.

Other pages this might help include Los Angeles p1407, p1467, p1499J.

danvk:
This was implemented by #168. Its merits vs. RANSAC remain to be sorted out.

PR #175, opened 2026-07-28
Streets-as-constraints georeferencer (issue #168): +1.5 aggregate score

danvk:
Fits a page from street labels treated as position+angle constraints against their named OSM polylines, with no intersection GCPs — issue #168's idea. Two distinct parallel streets fix rotation and, from their spacing, scale; one crossing street completes the pose.

**Measured on the twelve truth volumes: aggregate score 73.1 → 74.6 (+1.5), no volume regressing, disasters down from 3.2% to 2.8% of land.**

| volume | before | after |
|---|---|---|
| los_angeles_ca_1949_vol_14 | 78.1 | **82.6** |
| kansas_city_mo_1951_vol_4 | 67.8 | **70.7** |
| new_orleans_la_1896_vol_2 | 67.3 | **69.7** |
| grand_rapids_mi_1953_vol7 | 66.6 | 67.7 |
| brooklyn_ny_1939_vol_1 | 79.1 | 80.1 |
| washington_dc_1916_vol_2 | 78.4 | 79.3 |
| the other six | — | unchanged |
| **aggregate** | **73.1** | **74.6** |

Only eleven pages are adopted, but they are land-heavy — two LA pages carry +4.5 on their own.

## The solver

`mapsnap/street_solve.py` plus a truth-aware harness (`candidates | select | report | materialize`) and 22 synthetic tests. It reuses the region placer's math through three behavior-preserving extractions in `align_page_region.py` (`PoseSigmas`, `solve_pose`, `constraints_from_features`), so none of it is reimplemented.

It runs only where a page has a coarse location prior, restricts the vocabulary to that neighborhood, and is built around the measured fact that **a quarter to a half of assembled constraints are wrong** — renamed streets matching elsewhere, streets rerouted since the survey, near-square text boxes whose long axis is meaningless. Poses come from consensus over constraint pairs, inliers from hard angle/position gates, and a pose resting on a single constraint is rejected. Six anti-drift guards are enumerated in the module docstring; the re-snapping loop that made a predecessor version of this idea drift does not exist here by construction.

## What the pages taught it

- **A street whose OSM data stops short is not a misplaced page.** New Orleans p181's SOUTH MIRO sits **0.3 m from its street's line and 49 m past the end of it** — OSM draws less than the 1896 sheet did. Measuring point-to-segment charged that shortfall as position error, error *largest at the true pose*, so the solver slid pages away from truth to reduce it. Extending open polyline ends fixed p181 96 → 11 ft, KC p502 93 → 13 ft, Chicago p70w 376 → 11 ft.
- **ψ is mod-360, not mod-180** — the page has a top. Clustering bearings mod-180 averaged 25° and 205° into a meaningless 115°, which alone was the difference between LA p1499J abstaining and placing at 15 ft.
- **The unambiguous-only rule discarded same-street variants**: on p1499J every multi-candidate label was direction-prefix aliasing of one street, and merging them took constraint recall from 4 to 13.
- **Leave-one-out needs real redundancy** — at four inliers, dropping one leaves an exactly determined three that moves for sound reasons.

## Why it needs a referee, and which one works

On its own the channel is a coin flip against the incumbent: medians 13 ft vs 13 ft, head-to-head 162/150/400 across 712 pages. Worse, **nothing derived from the two channels can tell which is right** — their agreement predicts correctness at 50–52% across every disagreement band, and the solver cannot rank its own fits (inlier fraction is uninformative). An oracle that applies it only where the incumbent is bad is worth +81 pages, but the oracle needs truth.

`osm_snap.evaluate_pose` is the one signal derived from neither: road-skeleton chamfer against OSM plus name alignment. Scoring both poses with identical features picks the closer one **104 of 123 times (85%)**, and symmetrically — 79% when it prefers the street pose, 86% when it keeps the incumbent. `select` writes a sidecar only where it prefers the street pose by `ADOPT_GAP`.

The gate is forgiving: **every threshold from 0.05 up loses zero pages and creates zero disasters**, so it only trades how much upside is taken. 0.2 ships (91% precision).

Two things worth knowing for anyone extending this:

- **The incumbent is what the pipeline publishes, snap's sidecar included** — not the RANSAC fit beneath it. Snap already replaces that pose on 90 of the 155 RANSAC-vs-streets disagreements, so judging against RANSAC measures a fight production is not having and roughly triples the apparent headroom.
- **The existing 0.1 incumbent-verification gate is wrong for this challenger.** Where the incumbent is "defensible" (≥0.1), the street pose is closer only 19% of the time; the head-to-head gap is the signal, not a floor on the incumbent.

## Notes

`make_iiif_georef`'s page-key pattern needed `-streets` added: an unlisted variant produces no page key, so the glob matches the file and the annotation is silently short. Eleven adopted poses vanished that way and scored exactly as the baseline, which reads like a no-op rather than a bug. There is now a test per variant.

Measurements assume #176 (the edge-margin fix) — without it this channel silently drops a 2% band of page-edge labels that production keeps. Independent of #174; the referee call reads whichever feature form that branch has left in place.

778 tests, ruff and pyright clean.

### DIFF
--- a/mapsnap/keymap/align_page_region.py
+++ b/mapsnap/keymap/align_page_region.py
@@ -80,7 +80,7 @@ from mapsnap.keymap.locate import (
     page_number,
 )
 from mapsnap.keymap.page_regions import clean_cluster_mask
-from mapsnap.streets import Block, build_block_index
+from mapsnap.streets import Block, build_block_index, street_name_family
 from mapsnap.utils import FEET_PER_METER, haversine_m
 
 Point = tuple[float, float]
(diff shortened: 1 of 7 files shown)


danvk:
Los Angeles p1499J with RANSAC:

image

Los Angeles p1499J with streets:

image

Issue #12, opened 2026-05-15
Detect scale bar


danvk:
Every Sanborn map has a scale bar somewhere. Detecting and reading the text on these would give a direct way to determine the scale, without needing a street label or intersection.

Combined with #1, this would give another way to fit 1-gcp models.

This may be of limited value, though, since the scales for most of the maps in a volume are pretty similar.

[per the Library of Congress](https://www.loc.gov/collections/sanborn-maps/about-this-collection/#:~:text=noted%20as%20well.-,Interpreting%20Scale,-The%20vast%20majority):

> The vast majority of fire insurance maps were drawn at a scale of one inch to fifty feet (1:600) when expressed as a representative fraction. A smaller or less detailed scale of mapping was employed for suburban areas or large industrial sites. For these areas, scales of one inch to one hundred feet (1:1,200) or one inch to two hundred feet (1:2,400) were used. In rare instances, even smaller scales were used. Most map sheets contain a bar scale that facilitates the measurement of features or distances, and the scale is usually given in the title of the map as well.

danvk:
It would actually be very helpful to know, within a volume, which pages are expected to be at which scales.

For example, p21, p22, p23 and p24 in Champaign:

<img width="331" height="75" alt="Image" src="https://github.com/user-attachments/assets/4bfac516-aa8d-498d-8071-e90756e11185" />

<img width="344" height="52" alt="Image" src="https://github.com/user-attachments/assets/e2cd0bda-f891-4b4c-b4cb-67724720df72" />

<img width="335" height="67" alt="Image" src="https://github.com/user-attachments/assets/95ada350-2b14-468e-8784-100cf0762d46" />

<img width="315" height="55" alt="Image" src="https://github.com/user-attachments/assets/e3087f56-11c9-4102-810a-34b3c701bc53" />

They also have notes away from the scale bars in larger type:

<img width="233" height="60" alt="Image" src="https://github.com/user-attachments/assets/3b54c3da-0484-42f9-940a-9864782aac91" />

<img width="257" height="73" alt="Image" src="https://github.com/user-attachments/assets/9ede3262-5aa2-4fbe-8f4c-a7e086bfbd8d" />

<img width="254" height="66" alt="Image" src="https://github.com/user-attachments/assets/78a899ec-3189-417c-89b9-bb21e9d93199" />

[...]

PR #195, opened 2026-07-31
snap: rung-flip arbitration — adopt doubled-scale candidates over half-scale fits

danvk:
Closes the structural dead zone that let half-scale fits survive arbitration, using only candidates snap already generates. Follows the disaster characterization and rung-sweep prototype discussed on the 2026-07-30 run.

## The dead zone

Ten of the corpus's 39 disasters are fits at **exactly half or double their true scale** with near-perfect rotation. For several, the correct pose already sits in snap's candidate list — trapped between the two existing selection rules:

- **Refinement** requires <100 ft agreement with the incumbent; a correct-rung candidate disagrees by 200–700 ft.
- **Arbitration** requires an indefensible incumbent (verification < 0.1); a half-scale fit on a gridiron chamfer-matches well and verifies 0.14–0.89.

`rung_flip()` is a third rule for exactly this dispute: adopt a candidate at **1.7–2.3× the incumbent's scale** when it beats the incumbent's verification by **0.1**, doesn't lose the name head-to-head by more than 0.10, and carries a select score ≥ 0.9.

## The load-bearing gate is direction

**Up-flips only, never down.** Verification has a small-footprint bias — a half-scale candidate explains ¼ the ground and matches dense OSM cheaply. Calibrating over all **1228 rung-band candidate pairs in the twelve truth volumes**: every would-be down flip that passed any margin was a break (23–98 ft incumbents sent to 340–11,000 ft: NO-1951 p433/p434/p436, Chicago p61w, Detroit p21, Nashville p44, LA p1499o/r), while every fix but one was an up flip. A doubled candidate must explain 4× the ground, so verification preferring it is hard-won evidence.

Two referees were tried first and rejected with data:
- the **family-rung prior** endorses the wrong rung on exactly the contested pages (it was derived from the bad fits);
- the **volume median** breaks legitimate second scale families (Nashville's dual-scale corpus) and oversize sheets (LA p1499*).

## Validation (full snap → street-solve → iiif → compare on the five affected volumes)

| page | before | after |
|---|---|---|
| KC p556 | 406 ft | **13 ft** |
| DC p153 | 306 ft | **12 ft** |
| Nashville p4 | 396 ft | **26 ft** |
| Brooklyn p14 | 56 ft (streets rescue) | **15 ft** |
| NO-1896 p125 | 772 ft | 842 ft (disaster → disaster, neutral) |

Exactly one flip per volume — the calibrated set and nothing else — and **no other page moved anywhere**. A pleasing composition detail: Brooklyn p14's rung pose supersedes its street-channel rescue, and `street-solve`'s referee then correctly adopts nothing.

Corpus effect: ≥200 ft disasters −3, ≤25 ft +3 (~+0.5 land-weighted points). Brooklyn p57 (a *doubled*-scale fit at 1706 ft) is left untouched by design — fixing the down direction needs a verifier without the footprint bias (per-area normalization is the obvious candidate; noted for a follow-up).

815 tests (4 new, pinning the direction gate and each margin), ruff and pyright clean.

### DIFF
--- a/mapsnap/osm_snap_experiment.py
+++ b/mapsnap/osm_snap_experiment.py
@@ -1713,6 +1713,89 @@ def arbitrate_challenge(record: dict, arbitrate_gate: float) -> dict | None:
     }
 
 
+# Rung-flip arbitration: a fit published at HALF its true scale. Ten of the
+# 2026-07-30 corpus's 39 disasters were exact half/double-scale fits with
+# near-perfect rotation, and for several the correct pose already sat in the
+# candidate list — trapped between refinement (needs <100 ft agreement; a rung
+# variant disagrees by 200-700 ft) and arbitration (needs an indefensible
+# incumbent; half-scale fits chamfer well on a gridiron and verify >= 0.1).
+RUNG_UP_BAND = (1.7, 2.3)
+"""Candidate/incumbent scale ratios treated as an up-rung challenge.
+
+UP only, never down: verification carries a small-footprint bias. A half-scale
+candidate explains 1/4 the ground and matches dense OSM cheaply — calibrating
+over every rung-band candidate pair in the twelve truth volumes, every
+would-be DOWN flip that passed any margin was a break (23-98 ft incumbents sent
+to 340-11000 ft: NO-1951 p433/p434/p436, Chicago p61w, Detroit p21, Nashville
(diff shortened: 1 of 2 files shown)


danvk:
I hope #196 or a broader solution to #12 can obsolete this, but this is a win for now.

PR #196, opened 2026-07-31
ocr: admit the printed scale note into the rotation-0 vocabulary (#12)

danvk:
Implements the enabling half of #12: the printed scale note joins the constrained-decoder vocabulary, so OCR can finally *say* "Scale 100 Ft. to One Inch." Follows the feasibility study on that issue.

## The change

`SCALE_NOTE_TEXT`: the note phrases (with OCR-confusion variants, `IOO`/`I00`, following the pipe-annotation precedent) plus bar tick tokens. Three properties keep it contained:

- **Rotation-0 only** — the note and bar are always horizontal, so the vocabulary is unioned into the trie per angle pass and the rotated passes are untouched.
- **Ignore-tagged** — reads matching the set get `ignore: true` like the pipe annotations, so they never enter street matching.
- **No new pipeline stage** — it rides the existing recognition pass.

## Reliability: 173 of 173 numbers correct

Full recognition rerun over all twelve truth volumes (cached CRAFT boxes). Pages whose full note read at conf ≥ 0.2, with the parsed number validated against the page's truth scale:

| volume | full-note reads | correct | wrong |
|---|---|---|---|
| washington_dc_1916 | **110**/115 pages | 110 | 0 |
| brooklyn_ny_1939 | **50**/60 | 50 | 0 |
| hudson_co_nj_1950 | 8/89 (its 2× sheets) | 8 | 0 |
| champaign / detroit / NO-1951 | 3 / 1 / 1 | 5 | 0 |
| KC, LA, chicago, NO-1896, nashville, GR | 0 (bar-only or token styles) | — | — |
| **total** | **173** | **173** | **0** |

Notable single reads: Hudson p1 (a truthless pier page) reads its note at 0.91 — direct scale evidence for an unplaced page; Hudson p92 reads `SCALE 300 FT TO ONE INCH`, explaining its notorious 8.9% key-map region (3× scale = 9× area).

## Collateral: none at any confidence that matters

Isolated with a control rerun on unmodified code (which reproduced the pre-change outputs byte-identically — recognition is deterministic, so the diff is purely this change). On Hudson:

| confidence band | reads | outcome |
|---|---|---|
| ≥ 0.3 | 1077 | **1077 identical — zero collateral** |
| 0.1–0.3 | 846 | 831 same, 15 moved |
| < 0.1 | 9948 | 1662 junk reads now decode as ignore-tagged scale tokens |

The sub-0.1 churn is junk street-prefix reads (conf ~0.00–0.05) becoming explicitly-ignored scale tokens — noise labelled as noise. Verdict: **inline is clean; no separate pass needed.**

## Scope and follow-ups

This PR only makes the reads exist. Consumers are follow-up work: a `printed_scale_ft_per_in` parser feeding (a) the down-rung disasters #195 could not referee (Brooklyn p57's 1706 ft is a doubled-scale fit; a printed note settles the rung by authority), (b) the deferred 1-GCP scale, (c) scale-outlier gates. Coverage on bar-only volumes (GR/KC/LA) needs the geometric tick-spacing route from #12 — the tick digits are too small to read at 25%.

Evidence artifacts (per-volume tables, control diffs) in the session scratchpad; summary posted to #12.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,171 +23,125 @@ Test data comes from hand-geocoding by volunteers on OldInsuranceMaps.net:
 <table>
   <thead>
     <tr>
-      <th rowspan="2">Volume</th>
-      <th rowspan="2">Fit / Total</th>
-      <th colspan="3">RMSE Stats</th>
-      <th colspan="5">RMSE Distribution</th>
-      <th rowspan="2">Allmaps</th>
-    </tr>
-    <tr>
-      <th>Median</th>
-      <th>Mean</th>
-      <th>Max</th>
+      <th>Volume</th>
+      <th>Pages Placed</th>
       <th>≤25ft</th>
-      <th>25–50ft</th>
-      <th>50–100ft</th>
-      <th>100–200ft</th>
-      <th>200ft+</th>
+      <th>≥200ft</th>
+      <th>Score</th>
+      <th>Allmaps</th>
     </tr>
   </thead>
   <tbody>
     <tr>
-      <td><a href="https://oldinsurancemaps.net/map/sanborn03376_029">New Orleans 1951 Vol 5</a></td>
-      <td align="right">103/109 (94%)</td>
-      <td align="right">12ft</td>
-      <td align="right">18ft</td>
-      <td align="right">112ft</td>
-      <td align="right">84%</td>
-      <td align="right">10%</td>
-      <td align="right">5%</td>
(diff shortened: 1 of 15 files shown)


danvk:
This is inert for now, it just reads the "SCALE 100 FT. TO AN INCH" text.

PR #200, opened 2026-08-01
compare: score truth splits from GCP-verified selectors when panels.json is absent

danvk:
The `oim/pN.panels.json` split-region files need downloaded images (full-res `raw/` + OIM crops) that a fresh volume rarely has, and without them every placement of a truth-split page scores as unplaced. Since #190, selectors are repairable and — crucially — *verifiable*: this adds a fallback that derives panel polygons from the SvgSelectors, **gated per annotation on the frame test** (a selector must contain ≥50% of its own GCPs, else it may be crop-frame per OIM#402 and would grade the panel over its sibling's territory). Gated-out panels stay absent and score unplaced rather than wrong.

## Effect on the fresh Columbus 1951 vol 3 run

Nine truth-split groups had no panels.json; the fallback resolved **12 split-page rows** the missing-panels warning was silently zeroing:

| | before | after |
|---|---|---|
| placed | 86/105 (81.9%) | **98/105 (93.3%)** |
| score | 72.9% | **77.5%** (77.8% − 0.3%) |

p248's own panels pair across *disagreeing split numbering* — truth `[1]` = our `__2` at **8.0 ft** — exactly what the polygon-overlap matcher exists for. The small 0.3% disaster share that appeared is the bad-split sensitivity doing its job on a newly-scored panel rather than a regression.

The gate's negative case is pinned by a test: a frame-mixed annotation (GCPs in the full frame, selector in the crop frame) is refused, not misplaced.

Note: template-matched `panels.json` still wins when present (it's authoritative image geometry); the fallback only fills its absence. Scored-output vintage note: Columbus's `mapsnap.txt` on disk was regenerated with this + #199 combined.

823 tests (1 new), ruff and pyright clean.

### DIFF
--- a/mapsnap/compare_iiif_georef.py
+++ b/mapsnap/compare_iiif_georef.py
@@ -656,6 +656,36 @@ def redundant_skeleton_keys(truth_keys: set[str], generated_keys: set[str]) -> s
     return drop
 
 
+def selector_split_polygons(
+    truth_splits: list[dict],
+) -> dict[int, ShapelyPolygon]:
+    """Panel polygons from SvgSelectors, gated per annotation on GCP containment.
+
+    The fallback when ``oim/<page>.panels.json`` was never built (its inputs are
+    downloaded images). A selector is only trusted when it contains at least
+    half of its own annotation's GCPs -- the frame test from fix_truth_splits:
+    OIM's export writes some split selectors in the CROP's frame (OIM#402), and
+    a crop-frame polygon used here would grade the panel over its sibling's
+    territory. Panels whose selector fails the gate are simply absent, so they
+    score unplaced rather than wrong.
+    """
+    from mapsnap.fix_truth_splits import gcp_containment
+
+    polygons: dict[int, ShapelyPolygon] = {}
+    for item in truth_splits:
+        index = label_split_index(item)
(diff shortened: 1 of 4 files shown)


danvk:
Follow-up commit on this branch: **`mapsnap download-oim-splits`** — fetches a volume's OIM split crops + full-res parents by pure derivation (split pages and panel counts from `main.iiif.json` labels; the S3 URL prefix discovered by probing candidate volume slugs with state-name expansion, `columbus_oh` → `columbus_ohio`; `documents/`↔`regions/` layout reusing download-oim's swap downloader).

Ran the full loop on Columbus (9 split pages, 17 panels): download → `oim-split-truth` template matching → rescore. **The authoritative panels agree with this PR's GCP-verified selector fallback exactly — 77.5% either way, every shared panel identical to the decimal** (p248__1 8.0 ft, p244__1 896.2, p241__2 257.3). The image-derived truth also pairs two panels the selector path couldn't (p241__1 at 4.6 ft, p244__2 at 10.6 ft) — which nets zero score change here (both mid/good panels offset by rounding elsewhere), but shows the panels path remains strictly more complete.

That agreement is the best possible validation of the fallback's gate: where both paths can score, they produce the same answer.

PR #201, opened 2026-08-01
printed-scale: cold-start calibration + a note-derived rung in snap's ladder

danvk:
Breaks the printed-scale cold-start deadlock (#197's calibration needs note+fit pairs; some volumes print notes only on their unfittable odd sheets) and puts the note's rung into snap's search ladder. Motivated by Columbus p297 — a 200 ft district sheet carrying **13.9 of the volume's 22.5 missing points**, with its scale printed on the sheet at 0.92 confidence.

## Changes

1. **Median-rung calibration** (new middle tier): when self-calibration is impossible, assume the volume's median fitted scale is the 50 ft rung — `px_per_paper_inch = median_px_per_ft × 50` — guarded by a plausible-scan-resolution band (45–100 px/paper-inch ≈ 180–400 DPI raw) so a hypothetical 100 ft-median volume falls back rather than trusting a 2×-wrong rung.
2. **Corpus-default calibration** (last resort, volumes with no fitted pages): 62.5 px/paper-inch at the 25% working scale, logged as `corpus-default`.
3. **`printed-note` scale prior in snap's ladder**: the page's own printed scale, converted through the calibration, prepended when no existing prior is within 15% of it. p297's rescue previously searched *only* the volume-median 50 ft rung — the correct rung was never in the ladder.

## Why the median-rung tier: the corpus default was 21% off

Scoring p297 at its truth pose measured truth at **0.8066 m/px — almost exactly 4.03× the volume-median rung** (0.2001, the 50 ft sheets), while the 62.5 px/paper-inch default implied 0.9754, i.e. 3.8σ of the prior's sigma_log. Columbus scans at ~305 DPI, not ~250: the "LOC scanning is uniform" assumption behind the constant is false. Hudson is off the same way (median-rung 75.1 vs default 62.5). The median-rung estimate puts Columbus's note prior 0.8% from p297's truth scale.

## Columbus outcome — honest ledger

- The note check activated: p297's fresh refit came out at **1.42× its own printed note** and was correctly demoted (previously the check was silent on this volume). No other page moved. Score unchanged at 77.5%.
- **p297: the note rung was searched, and lost — and a controlled experiment shows why it cannot lock yet.** With the corrected rung in the ladder, the truth pose inside the search radius, and even a scale-appropriate P(road) (UNet re-run on the full-resolution scan), no candidate at the correct rung survives into the top 8. The full-res P(road) fixes the hallucination (16% → 0.8% of the page marked road) but on 200 ft-style line-drawn streets its **recall at the truth pose is 1.3%** — the model is style-blind, not just width-blind (bicubic-upscaled working images produce zero detections at all). Meanwhile wrong-rung small-footprint aliases verify at +0.44…0.61 in the dense grid.

So the note now delivers the right rung (at the right absolute scale) to every consumer, and the remaining wall for p297 is evidence, not search: a P(road) model that can see 200 ft-scale strokes, plus **per-area-normalized verification** (the same open item that blocks note-less down-flips). Either would let this prior place the single highest-value page in the corpus.

829 tests (4 new), ruff and pyright clean.

### DIFF
--- a/mapsnap/georef_from_labels.py
+++ b/mapsnap/georef_from_labels.py
@@ -441,7 +441,8 @@ def printed_scale_verdicts(
     if not notes:
         return {}
     calibration, calibration_source = resolve_px_per_paper_inch(
-        [(px, notes[img]) for img, px in scale_records if img in notes]
+        [(px, notes[img]) for img, px in scale_records if img in notes],
+        median_px_per_ft=statistics.median(px for _, px in scale_records),
     )
     if calibration_source != "self-calibrated":
         print(
--- a/mapsnap/georef_from_labels.py
+++ b/mapsnap/georef_from_labels.py
@@ -429,7 +429,7 @@ def printed_scale_verdicts(
     from mapsnap.printed_scale import (
         expected_px_per_ft,
         printed_scale_ft,
-        volume_px_per_paper_inch,
+        resolve_px_per_paper_inch,
     )
 
     notes: dict[str, int] = {}
@@ -438,11 +438,17 @@ def printed_scale_verdicts(
         note = printed_scale_ft(Path(stem + ".streets.json"))
         if note is not None:
             notes[img_path] = note[0]
-    calibration = volume_px_per_paper_inch(
+    if not notes:
+        return {}
+    calibration, calibration_source = resolve_px_per_paper_inch(
         [(px, notes[img]) for img, px in scale_records if img in notes]
     )
-    if calibration is None:
-        return {}
+    if calibration_source != "self-calibrated":
+        print(
+            f"Printed-scale calibration: {calibration:.1f} px/paper-inch "
+            f"({calibration_source})",
+            file=sys.stderr,
+        )
     verdicts: dict[str, tuple[str, float]] = {}
     for img_path, px_per_ft in scale_records:
         if img_path not in notes:
--- a/mapsnap/osm_snap_experiment.py
+++ b/mapsnap/osm_snap_experiment.py
@@ -568,6 +568,36 @@ def rotation_priors_for(
     return priors
 
 
+_note_calibration_cache: dict[str, tuple[float, str]] = {}
+
+
+def volume_note_calibration(vctx: VolumeContext) -> tuple[float, str]:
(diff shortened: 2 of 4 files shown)


danvk:
## ⚠️ Measured net-negative on two volumes — do not merge as-is

Refitting Columbus and Miami on this branch (tag `2026-08-01-adjacency`, with a no-adjacency control to separate a concurrent change) isolates the median-rung tier's effect:

| volume | before (corpus-default) | after (median-rung) | delta |
|---|---|---|---|
| Columbus | 77.5% | 64.5% | **−13.0** |
| Miami | 67.3% | 64.9% | **−2.4** |

**What happens.** The median-rung calibration is *more accurate* — and that is exactly the problem. p297's RANSAC fit sits at 0.6802 m/px, i.e. **0.84× its truth scale**, so it is genuinely consistent with the printed 200 ft note. The old, wrong calibration scored it at ratio 1.43 → `DROP (scattered)`; the corrected calibration scores it 1.18 → `KEEP (note-confirmed)`. The fit is then published — and it is **621 ft wrong in position** on a 2-GCP fit worth 13.9 points of land. The bad calibration had been an *accidental* guard.

The note check only validates **scale**, but a `keep` verdict also exempts the page from the relative-scale outlier heuristic that had been demoting it. Nothing else checks position. Miami shows the same shape more mildly (p42 22 ft → 1096 ft, p4 published at 3193 ft).

**Proposed guard.** A note-confirmed off-median fit should have to clear an *evidence* floor before publishing, not just a scale check. Measured on Columbus:

- p297's published (621 ft) pose: verification **−0.58**
- p297's truth pose: −0.48
- healthy (<25 ft) Columbus fits: median **1.36**, p10 0.51, min −0.46

So a floor anywhere around 0.0 rejects p297 while keeping every healthy fit sampled. Note the honest caveat from the null-normalization work: p297's *truth* pose verifies at −0.48 too, so verification cannot tell right from wrong here — it can only say **there is no evidence at this page at all**, which is precisely the case where a scale check alone must not license publication.

Recommend holding this PR until the guard is added and validated across the twelve-volume corpus; the calibration itself is correct and worth keeping.

PR #207, opened 2026-08-02
adjacency: one-sided claims as a lower-trust pose-graph tier

danvk:
Stacked on #206. Records unreciprocated-but-resolvable claims as a `one_sided` list in `adjacency.json`, and lets the **experimental** pose graph consume them via `posegraph --one-sided` (tagged measurements, 1.5× widened sigmas, `_1sided` output suffix).

> **Recommendation after Nashville's truth data: merge the data emission, do NOT promote the consumption.** The tier is far less precise than this PR originally claimed, and on one volume its pose-graph consumption is actively destructive. Details below. The `one_sided` field is inert in production (no production consumer reads it; `mapsnap fit` never calls `posegraph`), so the emission is safe and useful for future gated work.

## Corrected precision (this PR's original numbers were wrong)

The original description claimed "~90% genuine". That figure came from the **pre-#206** adjacency, where the union added 133 LA edges of which 118 were correct. It does not survive #206's repair layer, which raises claim recall (46→68% on LA) but lowers claim precision (93→79%) — and the extra claims land in exactly this tier. The Hudson "88%" was mis-derived from the union total rather than the added edges. Measured against the hand-labelled truth on the current generation:

| volume | key digits | claim precision | mutual edges | **one-sided edges** |
|---|---|---|---|---|
| Los Angeles | 4 | 79% | 125 @ **100%** | 193 @ **54%** |
| Hudson | 1–2 | 87% | 165 @ **100%** | 98 @ **49%** |
| Nashville | 1–2 | 63% | 79 @ **96%** | 165 @ **32%** |

One-sided precision tracks each volume's claim precision. Reciprocity is not a formality — it is the filter that removes junk reads which happen to resolve to a real page.

## Nashville (new truth: 70 keys, 311 references, all 1–2 digit)

- **vs label truth**: claims 63% precision / 69% recall; mutual **96%** precision (76/79), 49% recall; union 52% precision, 83% recall.
- **vs footprint-derived truth**: mutual **96%** precision, 38% recall; union 55% / 67%.
- **The two truth notions agree**, as on LA: 147/149 printed edges are footprint-adjacent (99%), while 53/200 footprint pairs (26%) are never printed — so `score_adjacency`'s recall denominator remains inflated by about a quarter.
- **#206's repair is provably inert here**: all 360 claim resolutions are `exact`, zero repairs — the direct confirmation that short-page-number volumes are untouched.
- **Every one of the 3 mutual edges contradicting the labels involves p8** (p2~p8, p49~p8, p50~p8 are junk reciprocation; p8~p29 is genuine), independently confirming the false-demotion diagnosis in #208.
- Misses: 14% gate loss (dominated by `height<28`), 11% CRAFT miss, 8% read/parse loss — unlike LA, where recognition was 49%.

## Pose-graph A/B on Nashville: the consequence

Same measurement file for both arms, so the comparison is exact.

| | placed | median | ≤100 ft | **≥200 ft** |
|---|---|---|---|---|
| mutual only | 5 | 72 ft | 3 | 1 |
| + one-sided | 10 | 637 ft | 2 | **7** |

**Five new disasters** (p6 2004 ft, p7 1304 ft, p8 1178 ft, p9 744 ft, p28 637 ft) plus p27 degraded 72→274 ft. At 32% precision the junk edges propose wrong pairs faster than verification and the Huber loss can reject them.

Full ledger across the four A/B'd volumes: LA +2 placements/0 disasters, Hudson neutral-positive (median 73→68 ft, disasters 5→4), DC +3 placements but p212 broken 55→266 ft, Nashville +5 disasters. That is not a feature to switch on by default.

## Caveats

The LA/Hudson/DC A/B numbers were measured before this branch's bare-number-fallback commit, which grew LA's tier from 111 to 193 edges; those arms were not re-run, so treat their deltas as indicative rather than current. Nashville's A/B is on the current generation.

843 tests, ruff and pyright clean.

### DIFF
--- a/mapsnap/edge_join_experiment.py
+++ b/mapsnap/edge_join_experiment.py
@@ -1837,10 +1837,11 @@ def measurement_sigmas(
     enter with loose sigmas and rely on the solver's Huber loss to be outvoted
     when wrong. Synthetic-anchor measurements (both pages unfitted, anchor
     posed at its keymap guess) searched a worse frame, so they get 1.5x;
-    measurements from one-sided-claim pairs (a printed reference whose
-    reciprocal never read — ~90% genuine against the label truth, but junk
-    reads land in the same tier) get the same 1.5x so mutual-edge evidence
-    outvotes them where they disagree.
+    measurements from one-sided-claim pairs get the same 1.5x so mutual-edge
+    evidence outvotes them where they disagree. That tier is only 32-54%
+    genuine against the label truth (vs 96-100% for mutual edges), so the
+    widening is doing real work — the join's own verification is the primary
+    filter, and the solver's Huber loss the backstop.
     """
     if verification >= 1.3:
         pos, theta = 5.0, 0.6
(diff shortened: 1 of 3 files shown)


danvk:
Corpus-wide adjacency regeneration (all 14 truth volumes) exposed and fixed a #206-stack regression: Chicago collapsed to 1 directed claim / 0 mutual edges because a bare printed "60" (referencing sheet p60w) resolved to nothing at *detection* time — the unique-number fallback existed only in `mutual_edges`, which never sees a read the resolver dropped. `resolve_page_key` now has the same fallback (`number` method: a digit read matching exactly one key's digit part resolves to that key; a digit part shared by many lettered sheets, like LA's 18 p1499*, stays ambiguous). Chicago after the fix: **425 directed claims, 127 mutual edges, 171 one-sided** — its densest graph ever. Commit 'resolve_page_key: bare-number fallback for unique lettered keys' on this branch; test coverage added.

PR #225, opened 2026-08-04
keymap-snap: place pages by matching P(road) against the key map (#211)

### LINKED ISSUE #211: Georeference by matching page ↔ keymap P(road) maps (aka keymap-snap)
For areas where the Sanborn-era streets no longer exist, matching streets with OSM is hopeless. #165 proposes joining adjacent pages by matching streets in their overlapping edges, which is a Sanborn ↔ Sanborn matching. This has proven a bit hard to pull off in practice, and it's also very sensitive to exactly how much overlap the pages in each volume have and how legible they are. Still, the idea of Sanborn ↔ Sanborn street matching is extremely appealing since it's impervious to urban renewal projects, highway construction, and other changes of the past 100 years.

Another Sanborn ↔ Sanborn idea is to match each page with its representation on the keymap. When the whole keymap process works correctly (page numbers identified, regions identified, keymap georeferenced), the fits are excellent. Here's Brooklyn p28, for example:

Image

So the locations can be quite precise. Here's the keymap for p45, where a highway has obliterated the grid:

Image

And here's the P(road) map for p45:

Image

Those should be very easy to match up, and this is completely resilient to changes in the underlying street grid. The georeferencing from the keymap drives the georeferencing for the pages.

A few things would need to be built to make this work:

1. Something to build a P(road) map from keymaps. This definitely _can_ be done, but the scale and color scheme means that the existing P(Road) UNet from #125 doesn't do well on keymaps.
2. Code to "snap" the two P(road) maps together. This should be a straightforward adaptation of the existing `mapsnap snap` code. This might need rotation and scale priors.

Depending on the quality of the keymap georeference, this might be more like a 25-50ft RMSE than <=25ft. Using polynomial2 or some more complex fit for the keyamps might help in that case, particularly on the edges (Chicago's keymap is ~1 block warped at the edges).

A motivating case for this is the missing "stack" of pages in Detroit, where a Stellantis factory has erased the historic grid:

Image

Here's that section of the keymap:

Image

The lack of internal streets for p7, p8, p22 and p30 are initially discouraging, but I think that is, itself a signal. There are streets on three sides of p7, and those give you your match:

Image

This could also unlock the elusive waterfront pages. For example, Hudson p92 is huge and has zero streets. But it has a complex waterfront shape that's marked with a blue border on the keymap:

Image

There's a matching blue border on the page itself:

Image

I'd need to train a new classifier to detect this waterfront marker. This wouldn't be v1 of this feature, but it's exciting because it's the first idea I've had that could potentially fit these pages.

I'm initially imagining this as a fallback technique for fitting otherwise-impossible pages but, if it works well, it could be used to validate existing fits from other sources. If a page's RANSAC fit is rotated or scaled differently than its keymap fit, that's suspicious.

danvk:
Closes #211

A key map draws the volume's whole street network on one georeferenced sheet, so every page is a rectangle somewhere on it. That makes it a placement channel for pages nothing else can reach — Detroit's 6/7/8/22/28/61/62 are all `nofit`, and this is the only thing that has ever placed them.

Eight commits: three that built the key-map P(road) UNet, four on the matcher, and one of Dan's own fixing a `centerlines.geojson` lookup for multi-volume sets.

`mapsnap keymap-snap <volume>` writes georef sidecars to `<volume>/kmsnap/`.

P(road) for Detroit keymap:

image

keymap-snap for Detroit p28, which is otherwise unplaceable:

image

## Results

Three volumes, each scored against truth and against **its own** centroid null — the same page placed on its key-map region centroid at the volume's median pose, no matching at all. The null moves with the key map's quality, so it has to be recomputed per volume rather than carried over.

| volume | median | vs its null | ≤50ft | ≤100ft | ≥200ft | placed |
|---|---|---|---|---|---|---|
| Detroit | **37.6 ft** | +198 ft | 59/93 | 79/93 | 10 | 95/102 |
| Washington DC | 53.7 ft | +212 ft | 54/109 | 78/109 | 24 | 116/136 |
| Miami | 124.0 ft | +87 ft | 9/64 | 26/64 | 27 | 66/91 |

Detroit's land-weighted score went **−19.4% → +5.4%** over this branch (−1.2% after the anisotropy fix, +2.7% after the rotation fixes, +5.4% after the GCP filter).

*(The "vs its null" column was measured before the GCP conflict filter, on medians of 39.0 / 64.7 / 145.1; the filter improves the snap side of each, so those margins are conservative.)*

## The design finding: the search space was the wrong shape

The matcher searches one scale and one rotation — an isotropic similarity — which is only correct in a frame where a correctly-placed page *is* one. Key-map georeferences are not that frame. Detroit's is **10.6% anisotropic**, so a page in its right place comes out stretched in key-map pixels and the correct answer sits outside the search space.

Measured by fitting the best isotropic pose to each truth pose and scoring it:

| page | p6 | p7 | p8 | p22 | p28 | p61 | p62 |
| best reachable | 25.7 | 25.5 | 27.5 | 25.0 | 26.5 | 17.8 | 3.7 |
| actually achieved | 69.6 | 52.6 | 64.4 | 42.7 | 41.2 | 53.8 | 43.6 |

**Median floor 25.5 ft against 52.6 ft achieved** — about half the error was structural, before the matcher made a single decision.

The fix warps the crop into a frame where the constraint holds. Polar-decompose the local linear part `L = R S` and warp by `W = S / sqrt(det S)`; then `L W⁻¹` is a scalar times a rotation, exactly the family being searched. `W` has unit determinant, so the crop is neither resized nor flipped, and the pose comes back through `W⁻¹`. No search dimension added, one small `warpAffine` per page.

Detroit, 93 pages with truth:

| | median | p75 | ≤25ft | ≤50ft | ≥200ft |
|---|---|---|---|---|---|
| sheet tangent, isotropic | 60.4 | 142.3 | 5 | 33 | 18 |
| page-local tangent, isotropic | 60.3 | 110.3 | 5 | 36 | 18 |
| **page-local tangent + stretch** | **44.0** | **68.4** | **21** | **56** | **14** |

The middle row is separable and worth reviewing as its own decision: a TPS has no single linear part, so the Jacobian is taken per page. That is a *tail* fix (p75 142 → 110); the stretch moves the median.

## Rotation: three latent defects, then neighbours' rotations

All three were invisible while every page in a volume sits at one angle, which is why Detroit's numbers looked fine and its foundations were not.

1. **Rotation was measured in lon/lat units.** A degree of longitude is cos(latitude) of a degree of latitude — 0.74 at Detroit — so the angle was wrong by ~6°, and the *sign* of the error flipped between a −26° page and a +56° one. It was also `atan2` of one matrix column, which equals the rotation only for a similarity; these maps are 10% anisotropic.
2. **The template angle was applied with the wrong sign.** The achieved rotation went as `2·crop − target`. Detroit's key map sits at −19° and its pages at −20°, so it landed 1.8° off — inside the ±3° jitter, and looked correct. Any volume whose key map is not aligned with its pages would have had every page placed at a systematically wrong angle.
3. **Rotating a template inside its unrotated bounds clips its corners**, and clips away most of a sideways page.

`angle_calibration` now measures the relation instead of deriving it: two probes determine a mapping known to be linear with slope ±1, and the slope is a free self-check. That check is what caught defect 1 — it came back −0.877.

With those fixed, neighbours' rotations become usable. Detroit's p65–p68 are sideways sheets, truth +56° against a −26° median. Swapping the volume median for the **neighbour median is worse than useless** — it fixes p67/p68 and breaks p63/p64/p69, because pages on the boundary of such a block have a mixed neighbourhood whose median is an angle no page has (>45° errors 4→2, but >10° errors 4→**5**). So neighbours only ever *add* a hypothesis, and the match score decides. Rescues: p65 646→64, p66 629→37, p67 641→84, p68 600→30, p32 340→69 ft.

## Contradictory GCPs

A key-map pixel matched to two different OSM intersections is a contradiction the spline can only average, bending the neighbourhood to do it.

| volume | inliers | conflicting | key-map spline residual |
|---|---|---|---|
| Detroit | 233 | 7 (3%) | 32 ft |
| DC | 137 | 10 (7%) | 55 ft |
| Miami | 507 | **99 (20%)** | **132 ft** |

`consistent_gcps` drops the whole group rather than picking a reading at random — nothing in the data says which is right. Miami 145→124, DC 65→**54** (≤50ft 36→54), Detroit 39→38.

## What generalization actually showed

**Page accuracy tracks the key map's own spline residual almost exactly:**

| | key-map residual | page median |
|---|---|---|
| Detroit | 32 ft | 38 ft |
| DC | 55 ft | 54 ft |
| Miami | 132 ft | 124 ft |

Miami is not a matcher failure. The matcher finds the right place on a drawing whose georeference is wrong. **The ceiling for any volume is set before the matcher runs.**

That also explains the match margin's behaviour. It is a usable quality gate on Detroit (≥0.10: median 35 ft, disasters 9→2) and DC (≥0.20: 54 ft, 24→2), and is **not** one on Miami, where high-margin matches are no better than low ones. Confidence is only meaningful where the key map is sound.

## Things tried and rejected, so they don't get retried

- **Chamfer refinement of the pose** (truncated, weighted, symmetric, stricter thresholds; translation / +rotation / +scale) — all six arms worse than baseline. Decisive reason: the objective prefers the *snap* pose over *truth* on **72 of 93** pages (2.19 px vs 4.00 px). Chamfer measures agreement with the key-map drawing; truth is agreement with the ground, and the key map's own error sits between them. Past a point, matching the drawing better means matching the world worse.
- **Thinning the key-map roads** to sharpen the rotation peak — sharpens it 6.5× (half-width 5.2° → 0.8°) but moves the argmax slightly *away* from truth. The wide objective's rotation peak is already within 1.43° of truth.
- **Less spline smoothing.** Swept 1e2–1e7. Driving the GCP residual to zero makes placements worse (disasters 9→16): the intersections are noisy and the spline is right not to chase them. `TPS_SMOOTHING = 1e6` is near-optimal on accuracy — though 1e5 loses on median and disaster count while *winning* the land-weighted score (4.2% vs 2.7%), because the pages it tightens carry more land. Left at 1e6; `--smoothing` exposes it.

## Known limitations

- **`TPS_SMOOTHING` is in source-coordinate units**, so its strength depends on sheet size. Tuned for this corpus's 5500–8400 px sheets.
- **Miami remains weak** (124 ft, 27 disasters of 64, score −33.3%). Its GCP construction is upstream of this PR.
- **DC's score is −22.0%** despite a 54 ft median: it places 116/136 pages, and the tail is expensive under a land-weighted metric.
- Three volumes, not the full corpus.

## Verification

- 918 tests pass; ruff and pyright clean. 26 unit tests in `snap_test.py`.
- Tests pin the things that broke: that `L @ inv(W)` really is a similarity, that the stretch never reflects, that `metric_theta` is immune to latitude *and* to anisotropy, that a requested rotation is the rotation achieved, that a mixed neighbourhood does not collapse to one meaningless average, and that a contradicted GCP is dropped rather than guessed.
- Two bugs were caught by end-to-end runs that all unit tests passed through: a translation composed in the wrong frame (median 41.7 → **195.2 ft**, rotation looking perfect while everything drifted), and `int(page.stem[1:])` crashing on Miami's `p10s` / DC's `p133s`. A third — `TPS_SMOOTHING` frozen as a default argument, making the constant unreachable — was caught only because a sweep returned byte-identical results across a 100× change.

Given that record, a coarse end-to-end assertion on Detroit in CI (median RMSE under a bound) would be worth more than more unit tests. (#227)

Note: some pose sets trip a GEOS crash in `mapsnap iiif`, fixed separately in #224. Independent of this PR, but you want both to build annotation pages reliably.


danvk:
Pushed cd557dd, which fixes three latent defects and then uses the room they leave.

**The PR's headline numbers move: median 41.7 → 39.0 ft, ≥200 ft disasters 14 → 9, land-weighted score −1.2% → +2.7%.** Five pages rescued, none regressed by more than 25 ft.

## What was wrong

All three were invisible while every page in a volume sits at one angle, which is why Detroit looked fine.

1. **Rotation was measured in lon/lat units.** A degree of longitude is cos(latitude) of a degree of latitude — 0.74 at Detroit — so the reported angle was off by ~6°, and the *sign* of the error flipped between a −26° page and a +56° one. It was also `atan2` of one matrix column, which equals the rotation only for a similarity; these maps are 10% anisotropic, so it drifted with the angle being measured.

2. **The template angle was applied with the wrong sign.** Composing through a y-down image and a north-up world negates it: the achieved rotation went as `2·crop − target`. Detroit's key map is at −19° and its pages at −20°, so it landed 1.8° off — inside the ±3° jitter, and looked correct. **Any volume whose key map isn't aligned with its pages would have had every page placed at a systematically wrong angle.**

3. **Rotating a template inside its own unrotated bounds clips its corners**, and clips away most of a sideways page.

Rather than re-derive the convention, `angle_calibration` now measures it: two probes determine a relation known to be linear with slope ±1, and the slope is a free self-check. That check is what caught defect 1 — it came back −0.877.

## Neighbour rotations

Detroit's p65–p68 are sideways sheets, truth +56° against a −26° median, unreachable by a ±3° ladder.

The obvious form of this — swap the volume median for the neighbour median — **is worse than useless**: it fixes p67/p68 and breaks p63/p64/p69, because pages on the boundary of a sideways block have a mixed neighbourhood whose median is an angle no page actually has. Measured: >45° errors 4→2, but >10° errors 4→**5**.

So neighbours only ever *add* a hypothesis. A corroborated, distinct cluster is searched alongside the volume median and the match score decides on image evidence. Rescues: p65 646→64, p66 629→37, p67 641→84, p68 600→30, p32 340→69 ft.

## One thing worth knowing about how this was verified

My first port of the fix passed all 21 unit tests and was catastrophically wrong end-to-end — median 41.7 → **195.2 ft**. I had applied the match translation in key-map space when `location` is found in warped-crop space, so every placement went through the stretch and the crop offset. Rotation looked perfect; everything drifted.

The unit tests didn't catch it because they check rotation and single-page recovery, not the composed frame. The full-volume run against truth did. `template_placement` now deliberately stops at crop coordinates so the shift can't be composed in the wrong frame, and says why in its docstring.

## Also worth flagging

Rotation is now **not** where the remaining error lives. Sweeping rotation with translation re-optimized at each angle, the shipped objective's peak is already within 1.43° of truth. Thinning the key-map roads sharpens that peak 6.5× (half-width 5.2° → 0.8°) but moves the argmax slightly *away* from truth — a sharper peak in a worse place. The residual is translation and scale, not angle.

912 tests pass; ruff and pyright clean.

danvk:
Description rewritten to cover the whole branch — it was Detroit-only and three commits out of date.

**Headline: it now runs on three volumes, and Detroit's land-weighted score is +5.4%** (from −19.4% when this branch started).

| volume | median | vs its null | ≥200ft | placed |
|---|---|---|---|---|
| Detroit | 37.6 ft | +198 ft | 10 | 95/102 |
| Washington DC | 53.7 ft | +212 ft | 24 | 116/136 |
| Miami | 124.0 ft | +87 ft | 27 | 66/91 |

Two bugs that only a second volume could expose:

- `int(page.stem[1:])` crashed on Miami's `p10s` and DC's `p133s`. Detroit's stems are all numeric, so no run and no test could reach it.
- **Miami's key map has 99 of 507 GCPs putting one pixel in two places**, median 233 ft apart. `consistent_gcps` drops the group rather than guessing — Miami 145→124, DC 65→**54**, Detroit 39→38.

The finding I'd most want a second opinion on: **page accuracy tracks the key map's own spline residual almost exactly** — 32/55/132 ft of key-map residual against 38/54/124 ft of page error. Miami isn't a matcher failure; the matcher finds the right place on a drawing whose georeference is wrong. The ceiling is set before the matcher runs.

That reframes the match margin too: a usable quality gate on Detroit (≥0.10: 35 ft, disasters 9→2) and DC (≥0.20: 54 ft, 24→2), and **not** one on Miami, where high-margin matches are no better than low ones.

The description now also records what was tried and rejected — chamfer refinement in six variants, road thinning, and less spline smoothing — with the measurement that killed each, so they don't get retried.

One process note worth acting on. Three defects in this PR were caught by end-to-end runs or by results that looked *too clean*, not by unit tests: a translation composed in the wrong frame (median 41.7 → 195.2 ft while rotation looked perfect), the lettered-stem crash, and `TPS_SMOOTHING` frozen as a default argument so the constant was unreachable. A coarse end-to-end assertion on Detroit in CI would be worth more here than more unit tests.

danvk:
This PR is good as a proof of concept. It's not active in the pipeline yet, but I'd like to start trying it on otherwise hard-to-place pages. When there are streets, the snapping works pretty well and gives us a new pathway to georeferencing that's limited only by the quality of the keymap georeference and page number detection.

PR #234, opened 2026-08-05
volume viewer: open page views inline, and fix their image paths

### LINKED ISSUE #130: Improve navigation in the IIIF viewer
You can click a link to go to a selected page's streets or georef view, but you can't easily go back to the IIIF page with your view state intact.

danvk:
Two changes to the volume viewer's page links: the feature that was asked for, and a bug it exposed in already-merged code.

Inline streets view:

image

Inline georef view:

image

Fixes #130 for real this time

## Inline views

Clicking **streets view** (or georef/adjacency) navigated away, losing the volume, page list and selection — so the workaround was ctrl-click and manage tabs. The view now renders where the map is, with the volume selector, page list and page info staying exactly where they were, and an **×** to dismiss it.

The **↗** beside each label is still a plain link with `target="_blank"`. This adds a way to stay put; it doesn't take away the way that already worked.

**Refactor**: `App` was 851 lines of debugger state with the volume-viewer branch at line 663. Rather than move 660 lines, it becomes a props-driven `DebugView` — `files` as a prop when embedded, `?files=` from the URL when standalone — under a ~15-line `App` that routes. `onClose`'s presence is what marks it embedded, since a standalone page has nothing to close back to and the two then can't disagree.

## The bug in #233

#233 built one `base` per page and used it for **both the image and the sidecar**. But an artifact directory holds a run's `json`/`txt` output and **never page images** — those exist once, at the volume root. So for any run that had saved sidecars, every link pointed at a jpg that doesn't exist. **The new-tab links were equally broken**; the inline view just surfaced it immediately.

```ts
const imageBase = `data/${volume}/${selectedPage.stem}`;
const sidecarBase = fromRun ? `${runArtifacts!.dir}/${selectedPage.stem}` : imageBase;
```

Two things about how it got through, since both are reusable lessons:

- I verified #233 by checking the endpoint's output and reading the constructed hrefs. Both were correct. I never followed one to see whether a page rendered — the endpoint being right and the links being wrong were independent failures, and I only tested the half I'd built.
- A `curl` status check would **not** have caught it: vite's dev server serves `index.html` for unknown paths, so the broken image URL returns **200**.

## Also

`.view-nav` is `position: fixed`, right for the standalone page's corner and wrong embedded — the close button landed 180px outside the container it closes, over the info panel. Now scoped to the standalone case.

## Verification

On the reported URL (`brooklyn_ny_1939_vol_1/2026-07-30-full`, p32): image loads from `data/brooklyn_ny_1939_vol_1/p32.jpg` at 1370×2067, sidecar from `artifacts/2026-07-30-full/p32.streets.json`, 9 detection rows, no error. Map hidden while open, restored with selection and URL intact on close. Standalone page keeps its "volume viewer" link, has no close button, and keeps its filter panel.

188 tests pass; tsc and prettier clean.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -9,6 +9,9 @@
 - Write unit tests for all public functions.
 - Functions should never take more than four positional arguments. To avoid this factor out dataclasses/interfaces, use keyword-only arguments or an options object, or group x/y parameters into  `(x, y)` tuples (or lat/lng, width/height, etc.).
 - All boolean variables (function parameters, command-line flags) should default to `false`.
+- Use American English:
+  - "Color" not "colour"
+  - "Penalize" not "penalise"
 
 ### Python
 
--- a/app/server/iiifAnnotations.test.ts
+++ b/app/server/iiifAnnotations.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it } from 'vitest';
 import {
+  imageStemsByLowercase,
   rescaleSvgSelector,
   rewriteAnnotationPage,
   serviceUrlToPageKey,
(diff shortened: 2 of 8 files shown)


danvk:
Fixed the zoom-and-reload on close.

**Cause**: the map was rendered in the false branch of a ternary, so opening a debug view *unmounted* it — and `VolumeMap`'s cleanup calls `map.remove()`. Closing remounted it, which re-ran the initial `fitBounds` (against a fresh `fittedVolumeRef`) and refetched the whole viewport. That's the jump and the tile reload.

**Fix**: the map stays mounted and is hidden instead, with the debug view laid over it rather than replacing it in the flex flow. It keeps its layout box, so nothing resizes either. Visibility is toggled on a wrapper rather than on `#map`, because the map sets its own inline `visibility` for its reveal — it only ever sets `hidden` or leaves it unset, so it inherits from the parent cleanly.

**Measured over a full open/close cycle**, after waiting for the initial tile load to settle:

| | |
|---|---|
| new tile requests | **0** |
| canvas instance | same |
| camera | `-74.00134,40.67916@14.19` → unchanged |
| selection | `p32` kept |

Worth noting my first measurement of this showed **90 new tiles** and looked like a partial fix. Those were the *initial* viewport still filling in — I'd sampled the baseline 4s after load, before it had settled. Re-measuring after waiting for the count to stabilise gives 0. A single before/after reading would have been misleading in either direction here.

Issue #250, opened 2026-08-06
_erase_underlines paints whole rows: leaves part of the bar and destroys digits (worse than no-op on Fargo p60__1)


danvk:
## Summary

`_erase_underlines` paints out **whole rows** of a detection box. Where a Sanborn ordinal underline shares rows with the glyphs above it — which is the normal case — this both leaves part of the bar and destroys part of the text. On Fargo p60__1 it is measurably **worse than doing nothing**.

## The mechanism

Rows in the bottom `scan_fraction` (0.25) with >= `dark_coverage` (0.25) dark pixels are set to white across the box's full column range. Two consequences:

1. **It misses the top of the bar.** The window is a fixed 25% of box height; on a 20px box that is 5 rows, and the underline starts 6 rows up. Fargo p60__1 "10TH": the underline occupies rows 899-901, the window starts at row 900. Row 899 is 0.27 dark — it would qualify — but is out of range. The residual 1px bar then **fuses the T and H**.
2. **It erases glyph pixels.** The bar is only under the `TH` (cols 23-45 of a 48px box), but painting the whole row also removes the bottoms of the `1` and `0` at cols 6-19.

```
10TH  898 ......##...###..###........#......##.....#......
      899 ......##...########.....#####################...   <- digits AND bar share this row
      900 ......##....######.....#######################..   <- production erases this row entirely
      901 .......................#######################..   <- and this one
```

## Measured effect (constrained decoder, the confidence the pipeline records)

| label | raw | production row-paint | stroke-only erase | TH stretched into the bar rows |
|---|---|---|---|---|
| 8TH | 0.1416 | **0.1243** | 0.9410 | **0.9787** |
| 10TH | 0.0037 | **0.0023** | 0.0299 | 0.0464 |

Production is **below raw on both labels**. Removing only the bar stroke takes 8TH from 0.124 to 0.94; stretching the `TH` to fill the vacated rows does slightly better still (0.98).

This matters concretely: 8TH is the only N/S street on p60__1 that could cross its three parallel avenues, so admitting it is the difference between zero RANSAC intersections and a fit. It currently fails the 0.15 confidence floor at 0.124.

## Why a parameter tweak is not enough

[...]

PR #263, opened 2026-08-07
Three-vote underline arbitration: raw / precise / legacy row-paint

danvk:
The 2026-08-06 baseline exposed a recall gap in the precise underline detector (#250/PR #260): it requires a rule run >= 12 px, but Nashville's squished ordinals live in 20x8 px boxes with 6-10 px rules. The eraser never fires there, and the raw crops read at 0.0-0.2 where the old row-painting eraser had been reading 0.6-0.9:

| read | 08-05 (legacy eraser) | 08-06 (precise) | this PR |
|---|---|---|---|
| nashville p3 `2ND` | 0.965 | 0.127 | **0.965** |
| nashville p60 `4TH` | 0.690 | 0.010 | **0.690** |
| nashville p10 `4TH` | 0.730 | 0.235 | **0.730** |

Baseline cost: **Nashville 58.0 → 41.4** (9 pages degraded), **champaign 97.8 → 90.1** (p20__1's `S NEIL` 0.476 → 0.265 slipped under the confidence-relaxed size floor).

## Fix

The two-read arbitration from #260 already makes aggressive erasure safe — a bad erase loses the vote to the raw read. So bring the legacy row-paint back as a **third candidate**: each affected box is read raw, precisely-erased, and legacy-row-painted, and `choose_best_read` keeps the highest confidence (ties prefer main, then raw, then legacy). Per box this is a strict superset of both the 08-05 and 08-06 behaviors:

- Fargo's large-rule ordinals keep their precise-erase wins (8TH 0.953)
- Fargo's BROADWAY-class damage keeps its raw protection (0.9997)
- Nashville/champaign's small-rule ordinals recover their legacy reads

`underline_removed` now flags any read where an erased variant won. Extra recognition cost lands only on boxes an eraser touches.

10 new/updated unit tests; 129 pass in detect_text + erase_underlines suites. Validated end-to-end on the three Nashville pages above (production OCR path, keymap-restricted vocab).

Note: the running 2026-08-06 baseline predates this fix — volumes with small-rule label styles carry the same signature there. A rerun once this lands will be the real baseline.


danvk:
Added the motivating fixtures to `testdata/erase_underlines` (4dd37e1), and they revise the diagnosis: the legacy vote covers **two distinct artifact classes**, not one.

1. **Small-rule ordinals** (`*_smallrule`) — the class described above: nashville p3 `2ND` (rule under min_run=12), p60 `4TH` (where the winning CRAFT box turns out to be the taller of three boxes over the same glyphs — the tiny 20x8 sibling hides the rule from the row scan entirely).

2. **Dash-line labels** (`*_dashline`) — champaign's `S NEIL` and nashville p10's italic `4TH` are **not underlined at all**: the labels are printed *on a dashed street line*, and the dash row crossing the bottom of the box is what the legacy row-paint strips. The precise detector *correctly* refuses these (ink below the run — they fail its underline definition), so no amount of rule-detector tuning reaches them; only the row-paint vote does.

S NEIL is verified end-to-end through the production OCR path on this branch: 0.476, `underline_removed: true`, matching 2026-08-05 exactly.

Each legacy fixture is also pinned to *differ* from the precise eraser's output, so if the two erasers ever converge on a case, a test flags it for revisiting.

Issue #184, opened 2026-07-30
Add an "isolate" feature to the volume viewer


danvk:
Relates to #130; sometimes it would be nice to see _only_ the selected sheet in the volume viewer, without having to go to the streets or georef view.

PR #268, opened 2026-08-07
viewer: resolve truth page keys from the label, not the service URL alone

danvk:
Fixes the volume viewer dropping most of a volume's truth annotations — reported as Grand Rapids showing "10 pages shown, 73 skipped" with `p844__3` greyed out as "not georeferenced" despite being present in `main.iiif.json`.

## Two bugs, one root cause

The TypeScript `serviceUrlToPageKey` took **only the service URL**. Python's `source_id_to_page_key` takes `(source_id, label)`. That divergence caused:

1. **Volumes that link no image service are entirely invisible.** 73 of Grand Rapids' 83 truth annotations carry `target.source.id: null`, so the derivation returned null and every one was skipped as `not-a-page`. (Python's docstring names this exact volume.)
2. **Split panels collapse onto their parent.** Only the label records the panel variant (`"... p844 [3]"`). Without it every panel of a sheet derives the same parent key, so multiple annotations claim one key and all but one is lost — **228 panels across the 15 truth volumes**:

| volume | panels lost | key collisions |
|---|---|---|
| fargo_nd_1958 | 55 | 33 |
| kansas_city_mo_1951_vol_4 | 30 | 15 |
| los_angeles_ca_1949_vol_14 | 26 | 14 |
| washington_dc_1916_vol_2 | 19 | 10 |
| columbia_sc_1956_vol_1 | 18 | 10 |
| champaign_ill_1915 / columbus_oh_1951_vol_3 | 17 each | 10 / 8 |
| (+8 more volumes) | | |

This is why `mapsnap compare` scored these pages correctly all along while the viewer could not show them — the Python path had both behaviours.

## Fix

Port the Python logic: `serviceUrlToPageKey(url, label)` appends the bracketed split suffix and falls back to a new `labelToPageKey` when the URL is null.

`iiifRoutes` builds its `localPages` map with its own call to the same function, so it must derive keys identically — otherwise `localPages` is keyed under names the rewrite never looks up and every page reports `missing-image` instead of resolving. (I hit exactly that in between the two edits.)

## Verified live

- **Grand Rapids: 71 of 83 kept** (was 10); `p844__1` and `p844__3` both resolve; no duplicate keys.
- **Fargo: 38 split keys resolved, zero collisions.**
- 14 new/updated unit tests covering split suffixes, the null-URL fallback, letter-only key-map sheets (`pa [2]` → `pa__2`), and labels with no page id. App suite 211 pass, `tsc` clean.

## Residual skips are real data gaps, not resolver failures

12 Grand Rapids and 17 Fargo truth panels have no local image because those sheets were never split locally (GR p725/p729/p819/p820/p834/p840/p841; fargo p7/p10/p12/p20/p43/p56/p58/p69). Both volumes do have `oim/*.panels.json` (16 and 22), so OIM splits some sheets our splitter leaves whole — worth a separate look.

### DIFF
--- a/app/server/iiifAnnotations.test.ts
+++ b/app/server/iiifAnnotations.test.ts
@@ -305,6 +305,85 @@ describe('rewriteAnnotationPage', () => {
   });
 });
 
+describe('rewriteAnnotationPage split panels', () => {
+  // A split panel is georeferenced against its PARENT sheet: the source is the
+  // parent's full-resolution size and the coords/selector are parent pixels.
+  // Serving the panel image rescaled them by the panel's aspect and rendered
+  // the sheet at ~2x scale, misaligned (fargo p45__1).
+  const parentSource = {
+    id: 'https://tile.loc.gov/…:06536_1958-0045/info.json',
+    type: 'ImageService2',
+    width: 6660,
+    height: 8070,
+  };
+  function panel(index: number) {
+    return {
+      id: `https://tile.loc.gov/…:06536_1958-0045__${index}/georef`,
+      label: 'Fargo, N.D. | 1958 p45 [2]',
+      target: { source: { ...parentSource } },
+      body: {
+        features: [
+          {
+            type: 'Feature',
+            properties: { resourceCoords: [3330, 4035], type: 'gcp' },
+            geometry: { type: 'Point', coordinates: [-96.8, 46.88] },
+          },
+        ],
+      },
+    };
+  }
+
+  it('serves the parent image while keeping the panel identity', () => {
+    const { annotation, skipped } = rewriteAnnotationPage(
+      { items: [panel(1), panel(3)] } as never,
+      new Map([['p45', { width: 1665, height: 2018 }]]),
+      'http://localhost:8182/iiif/fargo_nd_1958',
+      new Map([['p45', 'p45']]),
+    );
+    expect(skipped).toEqual([]);
+    const keys = annotation.items.map(
+      (item) =>
+        item.metadata?.find((entry) => entry.label === 'page')?.value ?? '',
+    );
+    expect(keys).toEqual(['p45__1', 'p45__3']);
+    for (const item of annotation.items) {
+      // Parent image, and coords scaled by the parent ratio (0.25) on BOTH
+      // axes -- the panel image would have given 1665/6660 vs 989/8070.
+      expect(item.target?.source?.id).toBe(
+        'http://localhost:8182/iiif/fargo_nd_1958/p45.jpg',
+      );
+      expect(item.target?.source?.width).toBe(1665);
+      expect(item.body?.features?.[0]?.properties.resourceCoords).toEqual([
+        832.5, 1009,
+      ]);
+    }
+  });
+
+  it('keeps a panel whose own image was never split out locally', () => {
+    // grand_rapids p729__2 has truth but no p729__2.jpg; the parent exists, and
+    // the parent is what the annotation is drawn against anyway.
+    const { annotation, skipped } = rewriteAnnotationPage(
+      {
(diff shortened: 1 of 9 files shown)


danvk:
Second commit (b9cb804) fixes a follow-on failure the first one exposed. With the truth annotations finally resolving, the map layer threw for every page under one tile wide:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'originalWidth')
```

`express-iiif` advertises no `tiles`, so `@allmaps/iiif-parser` synthesises a default tileset — and its fallback has an off-by-one:

```js
scaleFactors: Array.from({ length: maxExponent }, (_, exponent) => 2 ** exponent)
```

At `maxExponent === 0` that array is **empty**, so any image whose longest side is under the parser's 768 px tile width gets zero zoom levels and dies in `getTileImageRequest`. **22 split panels across the truth volumes are that small** — grand_rapids p844__3 (682×568), fargo p62__4 (365×488), champaign p2__1 (370×526), kansas_city p518__2 (322×749), los_angeles p1408__2 (455×458).

Fix: advertise a real tileset from our own service, which keeps the parser out of the fallback path entirely and is honest for a level2 service that serves any region and size. `withTiles` derives scale factors from the image's own dimensions and always emits at least `[1]`.

Verified live:

| page | size | advertised tiles |
|---|---|---|
| fargo p62__4 | 365×488 | `[{width: 512, scaleFactors: [1]}]` |
| grand_rapids p844__3 | 682×568 | `[{width: 512, scaleFactors: [1, 2]}]` |
| grand_rapids p714 | 1665×2018 | `[{width: 512, scaleFactors: [1, 2, 4]}]` |

…and a 512 px tile request against the smallest page returns a JPEG (HTTP 200). App suite 215 pass.

Worth reporting the `Array.from({length: maxExponent})` off-by-one upstream to allmaps/allmaps — even with tilesets advertised, any IIIF service that omits `tiles` will break their parser on sub-768px images.

danvk:
Third commit (02bfc94) adds #269 and #184, plus a regression fix from commit 2.

**#269 — skew and anisotropy in the info panel.** `distortion()` ports `truth_distortion` from `compare_iiif_georef.py` (same definitions, so the panel agrees with the `skew°`/`aniso` columns of `mapsnap compare`), working from the projected corners rather than the affine matrix. Values past 1° skew or 5% anisotropy render red, which is the signal that matters when auditing truth — a flat sheet photographed square cannot shear.

Verified on the page currently under inspection, grand_rapids p844__3: the panel reads **−1.13° / 0.678**, matching an independent affine fit through its 3 GCPs to two decimals. p714 reads 0.88° / 1.001 against compare's +0.92 / 0.994 (small difference expected: compare uses the annotation's own polynomial transform, the panel uses the corner-derived first-order behaviour).

**#184 — "Isolate selected".** Hides every warped image except the selected page's, so a sheet can be checked against the basemap without leaving the volume view. Implemented by toggling each map's `visible` option rather than re-adding a filtered annotation: `clear()` + `addGeoreferenceAnnotation` drops every map's fetched tiles and re-runs `fitBounds`, i.e. a full reload and a viewport jump for what should be a display toggle. Disabled when nothing is selected (isolating nothing would blank the map), a no-op for a missing page (no warped image), and round-trips through `?only=1`.

**Regression fix.** Commit 2 made the server's `page` metadata the on-disk stem, so a split panel already carries `__N`. `pagesFromAnnotation` appended the label's `[N]` a second time → `p844__3__3`, which matched nothing: every truth panel showed "not georeferenced" and its list row was greyed out. Caught by visual check, not by tests; there are now regression tests for both the split and whole-page cases.

App suite 221 pass, `tsc` clean, no console errors in the live viewer.

PR #300, opened 2026-08-11
Score every sheet equally, split by paper area

danvk:
## The problem

Weighting by absolute land **area** let a physically enormous low-density sheet outweigh a compact dense one. Hudson scores 50.5% while placing 76% of its pages:

| hudson | n | median land | total land |
|---|---|---|---|
| good | 62 | 0.058 km² | 3.78 |
| **unplaced** | **22** | **0.084 km²** | **2.82** |

Its unplaced sheets carry 1.45× the land of a typical placed one, so 24% of the sheets consumed 43% of the weight. The land fraction *was* discounting their water — it just could not stop a huge waterfront sheet outweighing a compact downtown one in absolute terms. The score was reporting sheet size as much as fit quality.

## The change

A page's weight becomes `sheet_portion × land_fraction`:

- **`sheet_portion`** normalizes each SHEET to weight 1, split across its panels by the paper area each occupies. A four-way split totals one sheet; a 95/5 split weighs 0.95/0.05. A volume can no longer gain or lose score by how finely it was cut.
- **`land_fraction`** is unchanged, and it is doing real work: without it brooklyn falls 9.3 points, because its unplaced sheets are genuinely mostly water. With it, 1.5.

Split portions come from `oim/pN.panels.json` — the OIM region boundaries `mapsnap compare` already matches splits with — **not** the truth `SvgSelector`, which is a clip mask and conflates masking with splitting (unsplit sheets' selectors cover only 0.69–0.79 of their source rectangle). All 122 split sheets across the 18 truth volumes have boundaries, and they tile their sheet almost exactly (median coverage 1.000); the equal-share fallback never fires.

The multi-volume aggregate is now a **simple mean**, not a pooled total: a volume is the unit of work, and pooling let the largest volumes decide the corpus number.

## Effect, same fits (last night's `pubab-reconcile`)

| volume | land-area | **new** | Δ | | volume | land-area | **new** | Δ |
|---|---|---|---|---|---|---|---|---|
| champaign | 98.0 | 97.6 | −0.4 | | chicago | 75.1 | 80.0 | +4.9 |
| new_orleans_1951 | 93.3 | 93.4 | +0.1 | | miami | 77.5 | 77.9 | +0.3 |
| columbus | 77.8 | **89.5** | **+11.7** | | detroit | 71.4 | 76.5 | +5.1 |
| brooklyn | 89.6 | 88.2 | −1.5 | | grand_rapids | 76.8 | 74.5 | −2.3 |
| los_angeles | 79.2 | **87.6** | **+8.4** | | columbia | 69.1 | 74.3 | +5.2 |
| philadelphia | 82.2 | 83.4 | +1.2 | | richmond | 73.4 | 74.0 | +0.5 |
| washington_dc | 81.3 | 82.7 | +1.4 | | **hudson** | **50.5** | **72.9** | **+22.3** |
| new_orleans_1896 | 83.3 | 82.0 | −1.4 | | fargo | 65.2 | 69.3 | +4.2 |
| kansas_city | 80.6 | 81.4 | +0.8 | | nashville | 64.7 | 66.7 | +2.0 |

**Corpus 76.9 (land-weighted) → 80.6 (mean of 18).** Fourteen volumes up, four down, none by more than 2.3. The spread compresses from 47.5 points (50.5–98.0) to 30.9 (66.7–97.6), and hudson stops being a lone outlier — the work now concentrates visibly in nashville, fargo, hudson, richmond and columbia.

## Before merging

- **Regenerate the README results table.** Its numbers predate this weighting and are not comparable; I added a note saying so rather than hand-editing figures that need a corpus run behind them.
- Historical reports (`docs/baseline.md`, `data/2026-08-10-rerun.md`) quote old-metric scores. They are dated records of runs, so I left them alone; worth deciding whether to annotate.
- One truth-data oddity surfaced: miami p62 has two panels each covering ~98% of the sheet (normalized to 0.492/0.492) plus two slivers. Handled correctly by normalization, but it looks like duplicated OIM regions and may deserve its own look.

## Verified

1,101 tests, ruff and pyright clean. The old `test_summarize_weights_by_land_not_pages` asserted the very behaviour being removed and was replaced by tests pinning the new contract: size must not decide anything, water still discounts, splits share one sheet, a 95/5 split weighs 95/5, and the missing-boundaries fallback announces itself. `mapsnap score` over all 18 volumes reproduces the prototype's numbers exactly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,11 @@ Test data comes from hand-geocoding by volunteers on OldInsuranceMaps.net:
   </tbody>
 </table>
 
-**Score** is the land-weighted share of each volume georeferenced to within 25ft, minus the share placed 200ft or worse — so a volume is rewarded for accurate coverage and penalised for confident mistakes. Pages left unplaced count against coverage but are not penalised as errors. RMSE is measured over a grid of points inside each page's own truth footprint (split sheets are graded per panel).
+**Score** is the weighted share of each volume georeferenced to within 25ft, minus the share placed 200ft or worse — so a volume is rewarded for accurate coverage and penalised for confident mistakes. Pages left unplaced count against coverage but are not penalised as errors. RMSE is measured over a grid of points inside each page's own truth footprint (split sheets are graded per panel).
+
+Each **sheet** carries the same weight, split across its panels by the paper area each occupies, and discounted by the share of the sheet that is usable land (approximated by proximity to an OSM street). So a sheet counts for no more because it is physically large, but a sheet that is mostly harbour counts for less than a dense downtown one.
+
+> **Note:** the table above predates this weighting and is not comparable to it; it will be regenerated from the next full corpus run.
 
 You can view the fits on Allmaps or get the IIIF files from the `gallery` directory. For notes on poor fits, see [test data notes].
 
--- a/mapsnap/score.py
+++ b/mapsnap/score.py
@@ -1,16 +1,39 @@
-"""Land-weighted success score for georeferenced volumes.
+"""Success score for georeferenced volumes: every sheet counts once.
 
-The project's success metric: **the share of truth land area on pages
-georeferenced to within a good-fit threshold (default 25ft RMSE), minus the
-share placed disastrously (default >=200ft)** — placing a page 2000ft off is
-worse than not placing it at all, so a disaster subtracts what a success adds.
-Unplaced truth pages stay in the denominator and earn nothing.
+The project's success metric: **the weighted share of truth pages georeferenced
+to within a good-fit threshold (default 25ft RMSE), minus the share placed
+disastrously (default >=200ft)** — placing a page 2000ft off is worse than not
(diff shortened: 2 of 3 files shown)


danvk:
The idea with the original score function from #148 was to de-emphasize small splits and waterfront areas. It succeeded at the former, less so at the latter. Hudson is the poster child:
image

It places 75+% of pages, but only scores 50% because the unplaced waterfront pages are enormous. The definition of "water" is a bit slippery (ha!) because of infill over the years. Hudson's p92 is not 50x more important than pages in the more densely developed areas.

This new metric is simpler and more uniform: each page is weighted equally. Small splits are downweighted by the fraction of their page that they occupy, so a small 5% split gets a 0.05 multiplier. If the waterfront areas were more important, the Sanborn surveyors would have surveyed them at the same scale as the downtown areas.

PR #322, opened 2026-08-14
Remove the underline erasure path (#264): the fine-tuned recognizer reads through the ink

### LINKED ISSUE #264: Street labels printed on dashed lines: no eraser removes them, and the row-paint vote can misread them confidently
Sanborn sheets print some street labels **on** a dashed line — the dashes run through the label at glyph height, not below it as an underline does. This is a distinct artifact from the ordinal underline handled by #250/#263, and nothing in the pipeline removes it.

## The evidence

`testdata/erase_underlines/champaign_sneil_dashline.before.png` (champaign_ill_1915 p20__1, `S NEIL`, angle 270) is the clearest case: the label reads `-S. -NEIL-` with dash segments interleaved between the glyphs. `nashville_pipe_dashline.before.png` is a second instance.

The legacy row-paint eraser (`legacy_erase_rows`, restored as an arbitration vote in #263) appears to help these — S NEIL reads 0.476 with it vs 0.265 raw — but **it does not remove the dashed line**. Measured on the fixture: the dash ink spans rows 7–18 of the 26-row box, while `scan_fraction=0.25` puts only row 18 inside the scan window. Exactly one row is painted white; the line survives. The confidence gain comes from clipping one arbitrary row, which is why the current behaviour is pinned in testdata as behaviour rather than a fix.

Neither eraser is designed for this. The precise detector (#250) refuses the box outright and correctly so — there is glyph ink below the run, which is exactly the guard that keeps `14TH`'s crossbar from being erased. Loosening it to catch dashes would reintroduce that regression.

## Why it matters

- **Systematic confidence suppression.** Every label printed on a dashed line is read against interfering ink. S NEIL at 0.265 raw sits close enough to the floors that it slipped the confidence-relaxed size gate in the 2026-08-06 baseline, which cost champaign 7.7 land-weighted points on one page (p20__1, 5.2 → 26.8 ft).
- **It manufactures false positives.** `nashville_pipe_dashline` is a `6" W. PIPE` annotation on a dashed line that the legacy variant misreads as **"4TH" at 0.730 confidence** (0.235 raw). That is worse than a low-confidence read: `4TH` enters street matching as a numbered street, whereas the pipe-label forms in `NON_STREET_TEXT` are excluded by design. Fargo's grid is full of real 4TH streets.

## Possible directions

1. **Detect and remove dashes properly** — a dashed line is a collinear run of short, evenly-spaced, equal-thickness components crossing the whole box. That is a describable structure, and unlike the underline case the removal should be inpainting at glyph height (harder: it overlaps the glyphs).
2. **Recognize the dash-line class and pass it as a hint** rather than erasing — e.g. admit these labels at relaxed floors, since a suppressed confidence on a structurally-explained artifact is different from a genuinely ambiguous read.
3. **Revisit the OCR approach generally.** EasyOCR + prefix-constrained CTC is being asked to read period cartographic type through printed interference it was never trained on. The two-read/three-read arbitration (#260, #263) is empirical compensation for that, and it is accumulating variants rather than converging. A fine-tune on Sanborn labels — the OIM harvest in #261 could supply the crops — would attack the underline, dash-line, and squished-ordinal cases together instead of one heuristic each.

Related: #250, #251 (systematic "1"/"0" downweighting in this font), #263, #261.

danvk:
Closes #264.

Recognition-only check over every `testdata/erase_underlines` fixture, stock vs fine-tuned recognizer, unerased vs erased crops:

| case | stock before | stock after | FT before | FT after |
|---|---|---|---|---|
| fargo_8th_rule | 8I4@0.42 | 8T4@0.29 | **8TH@0.998** | 8TH@0.999 |
| fargo_13th_rule | I30@0.36 | I3T4@0.17 | **13TH@0.939** | 13TH@0.901 |
| queens_36th_rule | 36T4@0.54 | 36TH@0.95 | **36TH@1.0** | 36TH@1.0 |
| queens_34th_smallrule | "34TH@0.25 | "34TH@0.25 | **34TH@1.0** | 34TH@1.0 |
| champaign_sneil_dashline (#264) | S.NEIE@0.76 | S.NEIL@0.63 | **S. NEIL@0.959** | S. NEIL@0.776 |
| nashville_4th_smallrule | LTH@0.36 | 4TH@0.44 | UITH@0.373 | ITH@0.982 (confidently wrong) |
| nashville_2nd_smallrule | 2ND@0.82 | 2ND@0.95 | **2ND@1.0** | 2ND@0.983 |

With the fine-tuned weights, erasure never improves a decoded text; on the #264 dashed-line case (which no eraser ever actually removed) it *reduces* confidence; and its one remaining behavioural effect converts a floor-droppable misread into a confidently wrong one. The three-vote arbitration (#250/#263), both erasers, and the `underline_removed` flag are removed; the viewer still renders the flag for historical streets.json.

Deliberately kept: the training-time underline/dash augmentation (`ocr_augment`, `train_street_recognizer`) — it is why the recognizer is robust — and the testdata fixtures, which just served as the eval set for this decision.

One separate finding filed into the margin: the FT model reads the nashville *pipe annotation* as `4TH@1.0` erased or not — #264's false-positive concern now lives in `NON_STREET_TEXT` filtering, not in erasure.


danvk:
Addressed: testdata/erase_underlines removed (33 files), the eraser-history comment in detect_text.py deleted, ocr_augment's docstring reworded to not cite the deleted path (its measured ranges are the durable artifact), and the augmentation review sheet now renders the synthesized section only. 76 tests green, zero remaining references.

Issue #324, opened 2026-08-14
Veto ~180° poses: 0 of 1,332 truth pages are printed upside-down


danvk:
Miami p18 is a disaster rotated ~180° — it swapped streets for avenues, which reads as upside-down street text on a north-up map. Corpus measurement of whether that pose class is ever legitimate: theta of every truth page's transform (page +x vs geographic east, y-flip accounted), all 18 volumes:

| band | pages |
|---|---|
| ~0° | 668 |
| ~45° / ~315° | 294 / 127 |
| ~90° / ~270° | 74 / 169 |
| **112°–247°** | **0** |

1,332 truth pages, zero anywhere near inverted — pages rotate freely to ±90° and to arbitrary grid-aligned angles (NO-1896), but Sanborn never printed a sheet whose text reads upside-down. The prior is not "unlikely", it is 0/1332.

Proposal: a hard implausibility veto on poses whose theta falls in the empty zone (page-frame text direction, NOT deviation-from-volume-median — so it is immune to the #256 multi-family problem):

1. `ransac_hybrid` seed scoring — kills the miami-p18 street/avenue alias at the source;
2. the same check wherever poses enter arbitration (snap's rotation ladder has admitted 180° flips before — the adjacency backfill work recorded p13 going 13→972 ft on one), or as a reconcile energy veto, which catches the class from every channel at once.

Validation: miami p18 recovers-or-unplaces; corpus A/B for collateral (expected none — the zone is empty in truth).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

danvk:
Corpus scan of published fits (live finals, 1,519 poses): **4 in the empty zone, every graded one a disaster.**

| volume | page | θ | source | graded |
|---|---|---|---|---|
| fargo | p63__2 | 180° | georef | 27,617.8 ft |
| detroit | p77 | 116° | snap:1 | 687.5 ft |
| fargo | p70__1 | 179° | georef | ungraded (no truth panel) |
| fargo | p63__1 | 180° | georef | ungraded |

(miami p18 absent only because its volume was mid-refit during the scan.) Two notes for the implementation: detroit p77 at 116° — a snap rescue candidate — argues for vetoing the full 112–247° zone rather than 180±30, and the 3-RANSAC/1-snap split confirms both proposed sites are needed. These four pages plus miami p18 are the validation targets: each should recover or unplace, with zero collateral.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #270, opened 2026-08-08
Consolidate the fit pipeline's gates and rescues into a pose-graph reconciliation


danvk:
The fit pipeline has grown into 15 reject paths and 4 resurrection paths spread across four sequential stages (inventoried gate-by-gate in `docs/fit-pipeline.md`, with a 2026-08-07 example page for each). Every stage makes greedy, irreversible decisions with local evidence, communicating only through sidecar renames — the state machine is append-and-outrank, and nothing ever weighs two candidate poses for the same page against each other. This issue proposes consolidating the gates, rescues, arbitrations and referees into one per-volume **pose-graph / factor-graph reconciliation**, and defines the migration path.

## Why: every recent disaster is an ordering artifact, not an evidence deficit

Documented cases where the volume jointly contained the right answer and the pipeline's decision *order* prevented combining it:

- **fargo p55**: stage 1's key-map-outlier gate rejected the *correct* RANSAC fit (the key-map location itself is ~2.9 km wrong); stage 3's rescue then published a *wrong* pose near that same wrong prior. Both sidecars sit on disk at the end — `georef-keymap-outlier.json` (right) and `georef-osm.json` (wrong, published) — and no step ever compared them.
- **fargo p64 (pre-#254)**: the adjacency gate demoted an 8-GCP, 11 ft fit on one biased edge measurement, *before* snap ran; snap then re-searched from the contradiction hint. The demotion was irreversible even though every other signal (its own GCPs, scale family, neighbors) supported the fit.
- **fargo p58__2 (pre-#259)**: RANSAC's two quadrant hypotheses tied within 0.9% and the anchored key-map prior sat unused at selection time — used only afterwards, as an executioner.
- **nashville p4**: tightening a stage-1 band changed which pages fed the family-rung estimator, which changed stage-3's refinement on a page no gate touched (26 → 402 ft). Hidden coupling through a shared statistic, invisible to any per-page reasoning.
- **`scale_corroborated_by_neighbors`**: an escape hatch *inside* a gate, discovered by surprise — the system has grown non-obvious control flow that even its author no longer expects (a scale outlier "recovering" via six agreeing neighbors, which then preserved a 22% misscale, the corpus's largest single disaster).

Under the graph formulation each of these is just evidence with a weight, and the failure mode disappears structurally rather than via another special case.

## What the experiment record already established (design constraints, not guesses)

| lesson | source |
|---|---|
[...]

PR #328, opened 2026-08-14
Snap searches from demoted poses without un-demoting them (#315)

### LINKED ISSUE #315: Let snap challenge demoted poses without un-demoting them
Found by the #270 phase-4 A/B (arm D, branch `scale-bands-retire`). Four pages the scale-band gate had **correctly** demoted got fixed the moment the demotion was removed:

| page | demoted (rescue path) | fitted (challenge/refine path) |
|---|---|---|
| richmond p353 | unplaced | **8.4 ft** |
| miami p76 | 353 ft | **4.4 ft** |
| detroit p56 | 329 ft | 57.5 ft |
| washington_dc p175 | 34.5 ft | 17.8 ft |

The mechanism: a demoted page's `fit_state` puts it on snap's **rescue** path, which searches from key-map/contradiction centers and failed on all four. A fitted page gets the **arbitrate/refine** path, which starts from the incumbent pose — and a mis-scaled RANSAC pose is evidently still a good search init, because refine fixed even richmond p353's 3.06× scale error.

So the gate's 20/20 demotion precision (see #314) and these four fixes are not in tension: the gate is right that the pose shouldn't *publish*, and wrong to also deny it the challenge path. The fix is to decouple those — let snap run arbitration/refinement seeded from a demoted pose while the demotion stands, with the result entering as a challenger for the arbiter to judge.

Same shape as the nashville-p8 finding for the adjacency gate (its demotion lets the rescue publish unopposed): demotion currently conflates "don't publish this" with "don't use this", and the second half costs us pages.

Arm-D data to validate against is in the `retD` tags and `scale-bands-retire`; a fix should recover p353/p76/p56/p175 **without** republishing the ~12 catches arm D also republished (miami p74 at 8,190 ft etc.).

danvk:
Closes #315. A demoted pose (misscale/outlier/contradicted sidecar with corners) now anchors snap's search context exactly as an incumbent does — center joins the search, rotation enters as a prior — while the demotion stands: candidates found from the seed face the unchanged rescue bars, stamp gate, and reconciler entry. Position and angle only, never scale (the scale is often why it was demoted; the matcher sweeps the volume rungs regardless).

## Validation (4-volume A/B vs a shared current-main control)

| volume | ctl | fix | target page |
|---|---|---|---|
| richmond | 77.6 | **78.1** | **p353: unplaced → 5.1 ft** (from its own 3.06×-mis-scaled pose — better than the issue's arm-D 8.4 ft, and without un-demoting anything) |
| miami | 77.8 | 77.8 | p76 unchanged — no longer a #315 case: the current pipeline publishes it *fitted* at 4,162 ft, rot_err −179.9° → it's a #324 (180° veto) case now |
| detroit | 70.4 | 70.4 | p56 unchanged — fitted at 330 ft (a slide), not demoted |
| washington_dc | 85.6 | 85.6 | p175 already 17.2 ft in control |

Zero regressions; none of arm D's ~12 catches republish (they'd need to clear the rescue bars as ordinary candidates, which is exactly what they fail).

## The two no-op rounds, for the record

Round 1 seeded an *existing* search — but p353 is misscale **with no keymap location**, so no context existed to seed (`no_keymap`). Round 2 moved the anchor into `build_page_context` — but patched the *panel* loader; base pages load through `edge_join_experiment.load_page_units`, whose near-identical block never populated the field. A loader test now pins both behaviors so the twin-path trap is fenced.

### DIFF
--- a/mapsnap/edge_join_experiment.py
+++ b/mapsnap/edge_join_experiment.py
@@ -258,9 +258,18 @@ def load_page_units(volume: Path) -> list[PageUnit]:
         keymap_centers: list[tuple[float, float]] = []
         keymap_radius = 0.0
         keymap_regions = None
+        demoted_affine = None
         if georef is not None:
             if state == "fitted":
                 gen_affine = page_world_affine(georef)
+            else:
+                # #315: a declined pose (misscale/outlier/contradicted) still
+                # anchors snap's rescue search; nofit-style docs have no
+                # corners and stay None.
+                try:
+                    demoted_affine = page_world_affine(georef)
+                except (KeyError, TypeError, ValueError):
+                    demoted_affine = None
             inlier_int = sum(
                 1 for i in georef.get("intersections", []) if i.get("inlier")
             )
@@ -279,6 +288,7 @@ def load_page_units(volume: Path) -> list[PageUnit]:
             truth=truth,
             split_truth=stem in split_truth_parents,
             gen_affine=gen_affine,
+            demoted_affine=demoted_affine,
             inlier_intersections=inlier_int,
             inlier_streets=inlier_str,
             keymap_centers=keymap_centers,
--- a/mapsnap/edge_join_experiment.py
+++ b/mapsnap/edge_join_experiment.py
@@ -103,6 +103,11 @@ class PageUnit:
     keymap_centers: list[tuple[float, float]]
     keymap_radius_m: float
     keymap_regions: list[list[list[float]]] | None = None
+    # A pose its channel produced and DECLINED (misscale/outlier/contradicted
+    # sidecar with corners). Never an incumbent; #315 uses it only to seed
+    # snap's search, because a demoted pose is measurably a good init (refine
+    # fixed richmond p353's 3.06x scale error from one).
+    demoted_affine: np.ndarray | None = None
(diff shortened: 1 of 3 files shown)


danvk:
Merge rationale, since the immediate score delta is small (one page, +0.5 richmond, ~+0.03 corpus):

1. **The class is about to grow.** The 18 measured volumes are polished — two years of gate-tuning already resolved most demotions. New-volume onboarding is where misscale/contradicted demotions concentrate (Asheville-class scale confusion), and a demoted page with no keymap prior is currently a dead end: p353's configuration is the preview, not the exception.
2. **It reduces the cost of keeping the gates we kept.** #314's case for keeping the scale bands was partly inflated by demotion also killing the search. Now a kept gate forfeits only publication-by-that-channel, not the page — which weakens the tension that made the phase-4 gate decisions hard, and is the #270 direction (channels generate, the arbiter judges) applied to declined poses.
3. **Measured risk ≈ zero**: candidates from the seed face the unchanged rescue bars/stamp gate/reconciler entry, arm D's catches structurally can't republish (they fail those very bars as ordinary candidates), and the loader test fences a twin-code-path trap independently worth having.

Structural bet, cheap premium, small immediate payout.

Issue #335, opened 2026-08-16
Failed-fit pages should offer their candidate GCPs as location priors


danvk:
snap rescue and street-solve require a location prior, and when RANSAC fails that prior comes exclusively from the keymap. But a failed RANSAC run still *computed* location evidence: candidate GCPs (matched street pairs crossing in OSM) whose world positions are currently discarded with the fit.

Measured on the 2026-08-14 baseline, over the 142 unplaced truth pages:

- 26 have candidate-GCP intersections; their cluster medians sit at p25 **124 m** / median **206 m** from the page's true location — **15 pages within 250 m, 18 within 1 km** (inside snap's search radius).
- Best cases: champaign p4__2 at **9 m** (3 matched names), grand_rapids p725__3 at 19 m, columbus p221__1 at 87 m (9 names). richmond p353 (the #315 page) and kansas_city p548 (disaster catalogue) also appear — independent confirmation of both mechanisms.
- Potential: 15–18 additional placements ≈ up to ~1 corpus point, discounted by verification conversion. Crucially it works **keymap-free**: schenectady's unplaced 100–114 block (keymap misses + sparse reads) is exactly the target class for new-volume onboarding.

Trust model is already established: like contradiction hints and demoted-pose seeds (#315), a wrong center fails verification and costs only search time. The p75 of 1.2 km is the aliasing tail the bars absorb.

## Design

1. **georef**: when a page ends unfitted, cluster its candidate GCP world positions (dedupe ~100 m, keep top clusters) and record them in the failure sidecar as `gcp_hints`. Also fixes half of #334 (keymap-less failures currently write no sidecar at all — now there is something worth writing).
2. **snap**: `build_page_context` appends `gcp_hints` to the search centers for rescue-state pages, exactly like contradiction hints. Pages with no keymap location get their ONLY centers this way.
3. **Phase 2 — parallel streets**: 116 of the 142 unplaced pages have matched names with no crossings (parallel streets, the street-solve signature). Their constraint is a corridor, not a point: perpendicular position is pinned to block scale (~50–100 m) but longitudinal position is ambiguous along the streets' mutual extent overlap (hundreds of metres to kilometres on long avenues). Since search centers are a list, sample the overlap interval at radius spacing — 2–4 centers convert the 1D ambiguity into ordinary multi-center search, and short/distinctively-named streets clip the interval sharply.

[...]

Issue #334, opened 2026-08-15
Pages with no channel sidecar get no georef-final (contract gap #2)


danvk:
Schenectady first run: p108 and the panels of p110/p113/p114 have **no georef-final.json at all** — the arbiter's every-page contract (a poseless final claims the page key) has a second gap beyond the superseded-parent one fixed in #323.

Mechanism: these pages are keymap-missing AND read-sparse, so georef writes nothing (the `nofit` sidecar is only written when there is a keymap expectation to record), no channel produces any sidecar, `build_nodes` skips pages with no hypotheses, and the final-writer iterates nodes only.

Fix: reconcile should emit `corners: null` finals for every unit in the volume context regardless of whether any sidecar exists — the units list already covers them. (Alternatively georef could always write a `nofit` sidecar, but the reconcile-side fix honors the contract at the layer that owns it.)

Found while diagnosing Schenectady's 100–114 block: NOT an OSM-buffer issue (100% vocabulary hit-rate on those pages' reads) — they're sparse outskirt sheets whose keymap misses deny them every prior-driven channel. Several have 5–11 confident vocab-matched reads, enough for street-solve once a keymap location exists, so keymap labeling is the unlock.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #336, opened 2026-08-16
Failed fits seed snap's rescue from candidate-GCP hints and neighbor stamps

danvk:
Phases 1 and 2 of #335: a page that could not be fit now seeds snap's rescue from whatever location evidence it does have — its own candidate GCPs, and its neighbours' printed claims. Validated: **richmond p353 places at 6.1 ft** (unplaced since onboarding), richmond 78.5 → **79.6 (+1.1)**, disasters unchanged.

## The chain

1. **Phase 1 — candidate-GCP hints.** A failed fit clusters its candidate-GCP world positions (single-linkage 100 m, top 3, largest cluster first — the mutually-consistent story outranks lone aliases) and records them as `gcp_hints` in the failure sidecar. Written on all three failure paths, including the debug-nofit path that writes its own sidecar, and *even for keymap-less pages* — narrowing #334's no-sidecar gap.
2. **Phase 2 — neighbour-stamp centres.** A neighbour's printed claim of a page, through the neighbour's published pose, is a world point on their shared boundary: median 178 m from truth for mutual claims (83 unplaced corpus pages), 416 m one-sided. This geometry previously existed only for gate-demoted pages via contradiction hints; now any unplaced page with placed claiming neighbours gets them as rescue centres, mutual first.
3. **snap**: both PageUnit loaders (the #315 twin-loader lesson) carry hints and stamps; `build_page_context` appends them to rescue search centres. Candidates from these centres face the unchanged rescue bars and reconcile entry.

## Both centre sources are scoped to pages with NO other search centres

Widening an *existing* search proved harmful even when the added candidates ranked worse individually: volume-energy arbitration is a joint optimization over the whole pool, so extra centres on nashville p6 (which already had keymap centres) re-ranked its selection from the correct 11 ft rescue to a 4×-scale 425 ft one — first via stamps, then again via phase-1 hints. Both are now restricted to centre-less pages, which is exactly the target class (schenectady's keymap-less 100–114 block; both richmond conversions are `no_keymap` pages), and leaves every existing search pool untouched.

## The bug the validation exposed (pre-existing, now fixed)

p353's rescue found the correct lock TWICE (6.2/9.9 ft, select 1.60/1.59) from hint clusters 250 m apart — and `distinct_margin`'s twin test abstained on margin 0.015 against the winner's own duplicate. Root cause in two layers: the test compared the ladder-rung `theta_deg` (search entry) rather than the REFINED pose's rotation — the two entries were 14.65° apart while their refined poses coincided. With pose-derived theta (and pose-centre separation made explicit), the margin is measured against the genuinely distinct rival: 0.015 → 1.076.

**Scoped to rescue.** An unscoped first run flipped nashville p6 (11 ft → 425 ft): the rung-theta twin collapse had been acting as an accidental brake on fitted-page challenges, and the arbitrate path's docstring names why it should stay — replacing a placed page is the risky direction, ties keep the incumbent. `arbitrate_challenge` and the selection path (scoped by `fit_state`) keep the legacy margin; rescue keeps the fix, which richmond p353's placement depends on. Rescues of *unplaced* pages that converge to one lock from multiple rungs may still flip on the next corpus run — confident locks are the likeliest beneficiaries.

## Honest scoping from validation

- The corpus scan's 26-page crossing pool overcounts: OIM panel-numbering swaps created phantom targets (champaign's "unplaced" p13__1 is truth's p13__2 at 7.9 ft), sliver panels carry negligible weight, and admission gates drop some scan-level matches before GCPs form (richmond p348: 7 matched names in the scan, zero pipeline GCPs). Realistic pool: a handful of base pages.
- The narrow centre-less scope deliberately trades reach for safety. Widening to pages that already have centres needs the joint-arbitration interaction understood first (#270), not a bar tweak.

### DIFF
--- a/mapsnap/edge_join_experiment.py
+++ b/mapsnap/edge_join_experiment.py
@@ -30,7 +30,7 @@ import json
 import math
 import statistics
 import sys
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from pathlib import Path
 
 import cv2
@@ -108,6 +108,10 @@ class PageUnit:
     # snap's search, because a demoted pose is measurably a good init (refine
     # fixed richmond p353's 3.06x scale error from one).
     demoted_affine: np.ndarray | None = None
+    # Clustered candidate-GCP world positions from a FAILED fit (#335):
+    # location evidence snap's rescue uses as search centers, recorded in the
+    # nofit sidecar. Empty for fitted pages.
+    gcp_hints: list[tuple[float, float]] = field(default_factory=list)
     anchor_truth: bool = False
     anchor_free: bool = False
     rmse_ft: float | None = None  # generated-vs-truth RMSE
@@ -278,6 +282,9 @@ def load_page_units(volume: Path) -> list[PageUnit]:
             keymap_centers = [tuple(c) for c in keymap.get("centers", [])]
             keymap_radius = float(keymap.get("radius_m") or 0.0)
             keymap_regions = keymap.get("regions") or None
+            gcp_hints = [tuple(h) for h in georef.get("gcp_hints") or []]
+        else:
+            gcp_hints = []
 
         unit = PageUnit(
             stem=stem,
@@ -289,6 +296,7 @@ def load_page_units(volume: Path) -> list[PageUnit]:
             split_truth=stem in split_truth_parents,
             gen_affine=gen_affine,
             demoted_affine=demoted_affine,
+            gcp_hints=gcp_hints,
             inlier_intersections=inlier_int,
             inlier_streets=inlier_str,
             keymap_centers=keymap_centers,
--- a/mapsnap/georef_from_labels.py
+++ b/mapsnap/georef_from_labels.py
@@ -99,6 +99,10 @@ class ProcessResult:
     rotation: float | None = None  # directed rotation in radians, set on success
     deferred: dict | None = None  # set when deferred (exactly 1 GCP, needs scale)
     nofit_written: bool = False  # process_image already wrote a --debug nofit sidecar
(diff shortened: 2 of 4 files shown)


danvk:
Final validated state (stamp generalization + scoping built out, per review discussion):

| volume | baseline | branch | detail |
|---|---|---|---|
| richmond | 78.5 | **80.8 (+2.3)** | p353 unplaced→**6.1 ft**, p348 unplaced→**12.7 ft**, p314 held |
| nashville | 71.2 | 71.2 | byte-identical after scoping |
| miami | 78.5 | 78.5 | byte-identical (p76's transient removal was a wide-scope accident) |

## The load-bearing lesson: search centers are not free evidence

The volume-energy arbitration optimizes jointly over the candidate pool, so adding centers to a page that already has them can flip its selection even when every added candidate ranks worse individually — nashville p6's correct rank-0 rescue (11 ft) lost to a 4×-scale 425 ft pose purely from pool composition, twice, until both prior sources were narrowed. **Both hints and stamps now serve only pages with NO other search centers** — fill vacuums, don't crowd rooms. That scope keeps every measured win (richmond's conversions are `no_keymap` pages where hints/stamps are the only centers) and leaves every existing search byte-identical. Widening existing pools safely requires making the energy arbitration robust to pool composition first — future work, not this PR.

Also in the branch, standing on their own: the `distinct_margin` twin-test fix (refined-pose rotation, not ladder-rung theta), correctly scoped to rescue with the challenge/energy paths keeping the conservative legacy margin.

The keymap-less target class (schenectady's 100–114 block) is precisely the fill-vacuums case — stamps/hints as sole centers — and validates there once its keymap or truth arrives.

Issue #333, opened 2026-08-15
street-solve has no incremental cache: 42% of corpus fit runtime for 19 published fits


danvk:
Stage timing from the 2026-08-14 baseline (18 volumes, #321 instrumentation):

```
stage shares of 170 CPU-minutes:
  snap          49%   (4,958s — WITH its incremental candidates cache)
  street-solve  42%   (4,289s — recomputed from scratch every run)
  georef         5%   reconcile 4%   gate+iiif 1%
```

Worst volumes: kansas_city 779s, miami 658s, philadelphia 444s, richmond 425s — each spent every run, for a channel that published **19 fits corpus-wide (1.2%)**. Snap costs about the same but delivers ~650 fits, and the only structural difference is that snap's `candidates.jsonl` skips pages whose inputs didn't change while `street_solve_run`'s `cmd_candidates` re-solves every keymap-prior page unconditionally.

Fix: give street-solve the same inputs-keyed skip. Its per-page inputs are exactly the #320 hash set — `streets.json` content, the page's keymap location, the incumbent pose it referees against (snap sidecar else georef), and gate/solver constants — so a per-page record in `artifacts/street_solve/candidates.jsonl` keyed on those (mirroring snap's mtime/fit_state checks) reuses the previous solve when nothing changed. The referee comparison (ADOPT_GAP vs the incumbent) must still re-run when only the incumbent changed, but the expensive part is the constraint solve, which depends only on reads + prior.

Expected effect: ~40% off every warm corpus run (99 min → ~60), independent of the general #320 framework. Together with the LA p1499 snap tail (#192, 630s), these two are the entire remaining runtime story: fixing both puts a warm 18-volume corpus run around 45 minutes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

danvk:
Closing: the premise here — street-solve as 42% of corpus fit runtime — was overtaken by #339 before anyone built the cache.

Recomputed the stage table from the 2026-08-28 baseline manifests (#360) the same way this issue did. The 08-14 side reproduces exactly (4,289 s street-solve, 4,958 s snap, 170.1 CPU-min), so the two runs are directly comparable:

| stage | 08-14 (18 vol) | 08-28 (20 vol) |
|---|---|---|
| snap | 4,958 s (48.6%) | 6,596 s (70.5%) |
| **street-solve** | **4,289 s (42.0%)** | **1,628 s (17.4%)** |
| georef | 473 s | 573 s |
[...]

PR #339, opened 2026-08-21
Street-solve: batch consensus scoring (9x on heavy pages, no semantic changes)

danvk:
Street-solve recomputed 97% of its runtime as 14M scalar `residuals_at` calls — profiled on Kansas City's 8 constraint-heaviest pages: 343s, with `consensus_poses` at 334s of it. The psi × scale × pair × line search proposes ~19k poses per page and scored each against every constraint one interpreted numpy call at a time.

Three changes (#333 discussion items 2–4; explicitly **no caching**), none touching gates, guards, or the objective:

1. **Batch scoring.** With bearing and scale fixed, a proposal only *translates* each label's world point (`pose_world_of` is base + t), so all proposals in a scale group score as one broadcast per constraint — `batch_point_to_segments` is the vectorized twin of `point_to_segments`. Proposal generation itself now uses a closed-form 2×2 solve over precomputed line equations (`line_equations`) instead of building numpy matrices per pair (1.2M `np.linalg.solve` calls gone).
2. **Dedupe before scoring.** Identical placements reached from different pair/line combinations merge on a 1 m grid before scoring, instead of being scored repeatedly and deduplicated at 20 m afterwards.
3. **No residual recompute.** `mean_position` reads the batch's position residuals instead of a second `residuals_at` pass over the inliers.

## Timing

| workload | before | after |
|---|---|---|
| KC's 8 heaviest pages (profiled) | 343 s | **38 s** |
| Kansas City full candidates (128 pages) | 779 s stage in 2026-08-14 run | **124 s** |
| Miami full candidates (91 pages) | 658 s | **126 s** |

## Equivalence

Verified full-volume against the archived 2026-08-14 candidates (street-solve code unchanged since, so they're an exact before-baseline):

- **miami: 91/91 statuses identical**; max candidate rmse delta 10.9 ft on one page.
- **KC: 127/128 statuses identical** — p483 abstains in both runs, just via `polish-runaway` instead of `unstable-leave-one-out`. Remaining deltas are tie-breaks between equivalent-cost attempts (5 `scale_source`/`psi_source` label flips, max pose move 1.3 m — inside the solver's own 20 m/0.01 dedupe tolerance; KC p523 improved 18.6 → 15.0 ft). Root cause of the ULP-level shifts: `base + t` addition order and Cramer-vs-LAPACK 2×2 solves.

New tests: `batch_point_to_segments` equals the scalar `point_to_segments` on random soups; every kept consensus proposal's inlier set matches the scalar `is_inlier` path. Full suite 1042 passing; live `candidates.jsonl` files restored to their archived state after verification.

### DIFF
--- a/mapsnap/street_solve.py
+++ b/mapsnap/street_solve.py
@@ -386,41 +386,56 @@ def dedupe_scales(
     return kept
 
 
-def propose_translation(
-    first: StreetSegments,
-    second: StreetSegments,
-    line_a: tuple[np.ndarray, np.ndarray],
-    line_b: tuple[np.ndarray, np.ndarray],
-    psi_deg: float,
-    log_scale: float,
-    size: tuple[int, int],
-) -> StreetPose | None:
-    """The pose placing two labels exactly on two given (non-parallel) street lines.
+def batch_point_to_segments(
+    points: np.ndarray, starts: np.ndarray, ends: np.ndarray
+) -> tuple[np.ndarray, np.ndarray]:
+    """(min distance, bearing of the nearest segment) for MANY points at once.
 
-    With bearing and scale fixed, a label lying on a known line is one linear equation
-    in the translation; two independent lines determine it outright.
+    The vectorized twin of ``align_page_region.point_to_segments`` — the same
+    clipped projection and nearest-segment bearing, evaluated for an (N, 2)
+    array of points in one broadcast instead of N interpreted calls. Returns
+    (distances (N,), bearings (N,) in degrees mod 180).
     """
-    point_a, direction_a = line_a
-    point_b, direction_b = line_b
-    normal_a = np.array([direction_a[1], -direction_a[0]])
-    normal_b = np.array([direction_b[1], -direction_b[0]])
-    matrix = np.array([normal_a, normal_b])
-    if abs(float(np.linalg.det(matrix))) < 1e-6:
-        return None
-    base = (0.0, 0.0, psi_deg, log_scale)
-    offset_a = pose_world_of(base, first[0], size)
-    offset_b = pose_world_of(base, second[0], size)
-    rhs = np.array(
-        [
-            float(normal_a @ (point_a - offset_a)),
-            float(normal_b @ (point_b - offset_b)),
-        ]
-    )
-    try:
-        translation = np.linalg.solve(matrix, rhs)
-    except np.linalg.LinAlgError:
-        return None
-    return (float(translation[0]), float(translation[1]), psi_deg, log_scale)
+    delta = ends - starts  # (S, 2)
+    length_sq = (delta * delta).sum(axis=1)  # (S,)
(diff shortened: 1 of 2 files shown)


danvk:
## Follow-up candidate: referee-first skip gate (analysis, not in this PR)

Investigated whether the constraint-heavy pages (the cost tail this PR optimizes) can be skipped outright, on the theory that lots of streets ⇒ RANSAC already did well.

**Inlier count is not a safe skip signal.** 16 of the 18 street-solve adoptions in the 2026-08-14 corpus sit on weak RANSAC fits (1–8 intersection inliers), but the two exceptions kill any threshold: miami p82 had **30 intersection inliers yet sat at 107.8 ft, and street-solve took it to 12.6 ft** — the channel's best conversion — and grand_rapids p820 (28 inliers, no truth) was also adopted. Skipping at ≥10 inliers would cut ~66% of the search-cost proxy and forfeit both.

**Incumbent referee verification separates perfectly.** Evaluated `osm_snap.evaluate_pose` on the incumbent for all 18 adopted pages plus the 4 heaviest strong-RANSAC pages per volume (72 pages):

| population | incumbent verification |
|---|---|
| 17 adopted pages (one had no score) | −0.79 … **0.73**, median −0.09 (miami p82: −0.13) |
| 72 heaviest strong-RANSAC pages | 0.03 … 1.85, ~2/3 above 1.0 |

Proposed gate: in the street-solve stage, build the page context and **evaluate the incumbent before solving; skip the consensus search when it verifies ≥ ~1.0**. This is the same referee `cmd_select` already runs, just reordered — and it's principled: adoption requires beating the incumbent by ADOPT_GAP, and no adoption has ever displaced an incumbent verifying above 0.73. At a 1.0 gate all 18 adoptions survive with 0.27 margin, and pages it declines to skip are the right ones (columbia's heavy downtown pages verify at −0.4…0.2, exactly where the channel deserves its shot).

**Expected performance effect (post-this-PR numbers):** the batched solver costs ~1–3 s on a heavy page, which is comparable to one referee evaluation, so the gate saves roughly **30–50% of the remaining ~2 min/volume stage** (it avoids the solve *and* the challenger-side evaluation on skipped pages; the incumbent evaluation is paid either way in cmd_select for posed pages). Pre-batching it would have been the headline fix; now it's a tail trim.

Caveats: the 0.73 ceiling is an empirical bound from 18 adoptions in one corpus run, not a theorem — the gate should log every skipped page (no silent caps), and the threshold deserves one corpus A/B before trusting it.

danvk:
Corpus validation: the 18-volume #341 A/B ran on top of this branch. The only suspected regressions (KC −1.8, DC −0.3) reproduced **exactly on unmodified main** under the same cold-cache protocol (#342) — cache-temperature variance, not this PR's tie-breaking. No street-solve adoption differs on any volume (KC/DC adoption sets byte-identical to 2026-08-14). The no-semantic-changes claim stands at corpus scale.

Issue #340, opened 2026-08-21
RANSAC runner-up poses: emit near-ties, let snap/reconcile arbitrate them


danvk:
RANSAC keeps a single argmax winner and discards every runner-up, but on gridded cities its objective is nearly flat across block aliases. Measured on miami (debug-georef, 86 truth pages):

- 80/86 pages have a distinct pose scoring >=70% of the winner
- p13: the TRUE pose (10 ft) scored **99.9%** of a 1,195 ft winner — ranks 0/1/3 are the same wrong alias
- p35: winner 99 ft; ranks 2–5 at ~97% are all <=12 ft
- Risk side: 26 pages with a good (<=25 ft) winner carry a >=200 ft alias in their near-tie set, so consumption must be gated (incumbent verification: no cross-channel adoption has ever displaced an incumbent above 0.73)

Phase 1 (implemented): emit top-4 distinct near-ties as `runner_up_poses` in the georef sidecar; snap consumes them as gated challenge/rescue *search seeds*. A/B on miami+nashville: zero regressions, small sub-band gains (p1 32.4→26.0 ft), scores unchanged.

Phase 2 (open, the p13 lesson): search-seeding cannot rescue the motivating page class. p13's demoted (misscale) pose anchors a rescue search whose `cluster_search_centers(0.75×radius)` swallows the runner-up center, and the NCC ladder's top-K is dominated by grid aliases even though the searched disc covers truth — while the true pose sits *in the sidecar*, needing only judgment. Runner-ups should enter as **candidate poses** (evaluate_pose + the full gate battery), which is exactly the #270 reconcile hypothesis-set design: every `runner_up_poses` entry belongs in the page's hypothesis set alongside georef variants and snap candidates.

danvk:
PR #341 now includes phase 2a: promotion past misscale demotions. When the scale-outlier pass would demote a winner, a rung-plausible near-tie (>=90% score) is refit via its forced seed pair and becomes the incumbent. miami p13: unplaced -> 10.2 ft; volume 78.5% -> 82.0% (+3.6). The remaining open item is candidate-injection into snap arbitration for pages whose near-ties fail the promotion bars (p80's 0.89 ratio, sub-0.9) — the reconcile hypothesis-set design.

danvk:
[...]

PR #341, opened 2026-08-21
RANSAC runner-up poses: emit near-ties, promote past misscale demotions, seed snap challenges

danvk:
RANSAC keeps one argmax winner and discards every runner-up — but on gridded cities its objective is nearly flat across aliases (miami p13: the TRUE pose scored 99.9% of a 1,195 ft winner). This PR emits the near-ties and puts them to work through two consumers, gated so healthy pages' candidate pools are never touched. Tracking: #340.

## The mechanism, in three parts

1. **Emission** — `ransac_hybrid` sinks every scored pose; `select_runner_up_poses` keeps ≤4 distinct near-ties (≥70% of winner score, >30 m or >1.5° apart, never upside-down per #324) in the sidecar as `runner_up_poses`, each carrying its corner quad, score ratio, and its seed pair addressed by **crossing pixels + geo candidates** (indices are tier-relative and a pixel can carry several world candidates — both discovered the hard way, see commits).
2. **Promotion (the headline)** — when the scale-outlier pass would demote a fitted page as misscale, it first checks for a near-tie (≥90% score) whose scale the volume's rungs accept, and **refits the page with that pose's seed pair forced** through the ordinary pipeline. The promoted fit carries its own coherent inliers/GCPs, becomes the incumbent, and snap *refines* it instead of rescuing around the wrong pose's center. A refit that fails or lands off-rung restores the original sidecar and falls back to the demotion.
3. **Gated challenge/rescue seeding** — fitted pages whose incumbent verifies below 1.0 (measured ceiling: no adoption has ever displaced an incumbent above 0.73) get runner-up centers + bearings as extra local-challenge seeds; rescue-state pages get them alongside the #315 demoted seed.

## A/B (miami + nashville, cold caches, PYTHONHASHSEED=0; `runnerup-ctl` vs `runnerup-ab5`)

| | control | full treatment |
|---|---|---|
| **miami** | 78.5%, 82/93 placed | **82.0%, 84/93 placed (+3.6 net)** |
| **nashville** | 71.2% (= 2026-08-14 baseline) | 71.2%, zero page moves — five arms, never perturbed |

miami's moves: **p13 unplaced → 10.2 ft** (promoted at 99.9% score ratio — the page that motivated the channel, previously stuck as a misscale rescue searching the wrong block), **p76 352 → 4.5 ft** (a disaster converted: promoted neighbor p74 flips p76's snap arbitration; stable across two arms), p74 unplaced → 118.8 ft (band-neutral coverage), p1 32.4 → 26.1, plus sub-ft jitter. Worst regression anywhere: p24 +0.7 ft.

## Why seeding alone wasn't enough (the intermediate result)

Runner-up centers seeded into the rescue search could not save p13: `cluster_search_centers` merges centers within 0.75×radius, and the NCC ladder's top-K over the merged disc is dominated by grid aliases even though it covers truth — the pose needed *judging*, not re-deriving. Promotion sidesteps the ladder by refitting the exact pose; full candidate-injection into arbitration remains #340's reconcile-phase item.

Tests: emission selection (distinctness, upside-down, cap, pair addressing), promotion selector (rung check, near-tie bar, missing-pair skip), consumption gate (ver ≥1.0 untouched, dedupe, no-score abstention), loader round-trip. Full suite 1045 passing; ruff/pyright clean.

### DIFF
--- a/mapsnap/edge_join_experiment.py
+++ b/mapsnap/edge_join_experiment.py
@@ -321,9 +321,7 @@ def load_page_units(volume: Path) -> list[PageUnit]:
             keymap_centers=keymap_centers,
             keymap_radius_m=keymap_radius,
             keymap_regions=keymap_regions,
-            runner_up_affines=runner_up_affines_of(
-                georef if state == "fitted" else None, width, height
-            ),
+            runner_up_affines=runner_up_affines_of(georef, width, height),
         )
         if truth is not None and gen_affine is not None:
             unit.rmse_ft = grid_rmse_ft_between(
--- a/mapsnap/edge_join_experiment.py
+++ b/mapsnap/edge_join_experiment.py
@@ -30,7 +30,7 @@ import json
 import math
 import statistics
 import sys
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from pathlib import Path
 
 import cv2
@@ -86,6 +86,30 @@ class TruthFit:
     transform_type: str
 
 
+def runner_up_affines_of(georef: dict | None, width: int, height: int) -> list:
+    """Affines for a fitted sidecar's runner_up_poses entries (empty if none).
+
+    Shared by both PageUnit loaders (this module's and osm_snap_experiment's
+    panel loader) so the runner-up channel cannot fall out of sync between
+    base pages and split panels.
+    """
+    from mapsnap.road_model import page_world_affine
+
+    if not georef:
+        return []
+    affines = []
+    for pose in georef.get("runner_up_poses") or []:
+        try:
+            affines.append(
+                page_world_affine(
+                    {"corners": pose["corners"], "width": width, "height": height}
+                )
+            )
+        except (KeyError, TypeError, ValueError):
+            continue
+    return affines
+
+
 @dataclass
 class PageUnit:
     """One base (unsplit) page of a volume with everything the harness needs."""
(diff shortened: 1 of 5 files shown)


danvk:
## 18-volume corpus A/B (final, after dropping the seeding channel)

Treatment = this branch + #339's street-solve batching, cold caches, PYTHONHASHSEED=0, tag `runnerup-corpus2`; baseline = the 2026-08-14 archives. **Mean 82.0% → 82.3%.**

| volume | delta | driver |
|---|---|---|
| miami | **+3.6** | p13 promoted → 10.2 ft; p74 promoted → placed; p74's placement cascades p76 352 → 4.5 ft |
| detroit | **+2.2** | p56 promoted at 97.3% score → 330 → 24 ft |
| columbus | **+0.8** | p244 promoted: split panel 2 → 11.5 ft, panel 1 wrongly covered (new 896 ft disaster) — net positive, and the sheet-equal metric weighs the split honestly |
| champaign / fargo / LA / grand_rapids | +0.6 / +0.4 / +0.4 / +0.1 | disaster→abstention flips, variance-shaped (see below) |
| kansas_city / washington_dc | −1.8 / −0.3 | **not this PR**: reproduced exactly on unmodified main under cold caches (#342) — cache-temperature variance; the warm 08-14 numbers were one arbitrary resolution of pool-sensitive split-panel selections |
| other 9 volumes (incl. nashville, richmond, brooklyn) | ±0.0 | byte-stable |

**Promotions: 7 fired across 5 volumes, zero harmful.** The two that moved bands moved them the right way (miami p13, detroit p56); DC and NO-1896 promotions landed in-band.

**Why the seeding channel was dropped** (`cea99b2`): the first corpus pass attributed every runner-up regression to it — richmond p363 (23→34 ft), p380 (19→64 ft), brooklyn p44 (175→456 ft, new disaster) — each a challenge search perturbed under an incumbent that verified badly (0.04!) but was actually good. Across ~120 seeded pages it produced zero band gains of its own. With seeding stripped, richmond and brooklyn returned to exactly their baseline scores. Poses now reach the pipeline only as whole-fit incumbents via promotion; anything weaker belongs in #340's reconcile hypothesis set.

Attribution runs archived under tags `ss-only` (KC+DC on pure #339) and `main-cold` (KC on unmodified main).

Issue #27, opened 2026-05-25
Handle divided highways better


danvk:
Before #23 I averaged intersections, which often worked better for divided highways than the new approach, which keeps them separate (Sanborn street labels are usually in the middle of the divided highway).

Examples where this is a problem:

- new_orleans_la_1951_vol_5 p454
- Lots of intersections with E. Jefferson Ave in detroit_mich_1929_vol_11, e.g. p44, p50, p71

For New Orleans p454, it picks one intersection on the left of the Napoleon Av and the other on the right. This gets the scale close to correct, but it introduces a huge torque. Putting both intersections in the middle of the highway would work better:

<img width="1103" height="430" alt="Image" src="https://github.com/user-attachments/assets/9a53d61d-d9ad-4639-ba5c-e6cc4e2fad4b" />

danvk:
This shows up clearly in the New Orleans 1951 overview. The line of brown "medium-bad" fits down the middle is all Napoleon Ave:

<img width="955" height="1107" alt="Image" src="https://github.com/user-attachments/assets/cc9724ab-9dae-4263-81b3-77ad2e915415" />

danvk:
Miami p82 is another good example: mapsnap produces one GCP from each side of the divided highway, which produces a rotated fit.

danvk:
Surveyed the current behavior (2026-08-22) ahead of a centroid experiment.

## What the code does today

There is a hidden averaging band. Per label pair, `find_intersection_gcps` clusters the two streets' shared OSM vertices at **60 ft** (`_cluster_geo_coords`, single-linkage) and emits one GCP per cluster **at the cluster centroid (mean)** — so a narrow divided road already averages. At ≥60 ft the carriageways become separate world candidates at (nearly) the same pixel crossing (`SAME_PIXEL_TOL_PX`); RANSAC seed pairs pick ONE, and the centroid across clusters is never offered. Same-pixel alternatives can't seed together (degenerate 2×2), a page whose GCPs collapse to one pixel defers to the 1-GCP path (which already "tries each world candidate"), and straightening is disabled for multi-cluster junctions.

## Scale (18-volume corpus, #327 counters)

- **3,147 divergent crossings kept; 112 inside winning seeds** — 112 published fits rest on an arbitrary pick between world alternatives.
[...]

PR #344, opened 2026-08-22
Divided roads: offer the carriageway centroid as a crossing candidate (#27)

danvk:
Stacked on #341 (the runner-up emission is the harm-mitigation this depends on). Survey and analysis on #27.

## Change

When a label pair's shared-vertex clusters span ≤250 ft, their joint centroid joins the per-cluster candidates as one more same-pixel alternative — `centroid=True` on the GCP and in the sidecar's intersections rows, with `centroid_candidates_emitted` / `centroid_in_winning_seed` counters (the #327 pattern). The candidate loop is factored into `_pair_crossing_candidates`, shared by both emissions. Unit tests cover the divided-road fixture (~95 ft carriageways → 3 same-pixel GCPs, centroid at the exact midpoint) and the span cap (~315 ft → none).

## A/B (NO-1951 + detroit + brooklyn, cold caches, PYTHONHASHSEED=0, control = `runnerup-corpus2`)

| volume | emitted | in winning seed | net | pages |
|---|---|---|---|---|
| detroit | 143 | **4** | 78.6% = 78.6% | p42 18.5→20.0 only |
| brooklyn (jog risk) | 16 | 1 | 87.7% = 87.7% | p30 5.2→10.1 only |
| NO-1951 | 25 | 0 | 93.4% → **92.4%** | p485/486/487 improved ~2 ft each; **p531 15.2 → 94.3** |

The five centroid-seeded winners produced no band changes — where the centroid actually wins, it's fine, and the three NO sub-band improvements are the divided-road precision story. The cost is one page, and it's instructive:

- **p531 is pure pool perturbation**: 14 centroids entered its 61-GCP pool, *none* won the seed, but the enumeration change flipped RANSAC's argmax between near-ties (published fit 15.2 → 94.3 ft).
- **The old pose did not survive in `runner_up_poses`** — at ~29 m from the new winner it falls just inside the 30 m distinctness floor and merged away. The floor is tuned for block aliases; divided-road/near-tie perturbations live at 10–30 m, *below* it. That's the concrete tuning question this experiment surfaces for #340: either a lower floor (~15 m), or always retaining the displaced previous-best, would have preserved exactly the pose the reconciler needs.

## Recommendation

The counters and the `centroid=True` plumbing are unconditionally useful (they make the 3,147-crossings/112-winning-seeds phenomenon visible per run). Whether to let the candidate into the pool *before* the reconciler exists is a judgment call: today it's net −0.3 on the three test volumes via one preventable casualty; with the distinctness-floor fix or reconcile-era judging, the analysis says it becomes safe. Happy to gate emission behind a flag, or fix the floor, before any corpus run — reviewer's choice.

### DIFF
--- a/app/src/iiif/compare.test.ts
+++ b/app/src/iiif/compare.test.ts
@@ -51,3 +51,12 @@ describe('statsByItemIndex', () => {
     });
   });
 });
+
+describe('statsByItemIndex casing', () => {
+  it('pairs uppercase lettered stems with lowercase sidecar keys', () => {
+    // compareTxt normalizes keys to lowercase; disk stems keep their case
+    // (asheville p3C). The join must not lose those pages' colors.
+    const stats = statsByItemIndex([row('p3c', 12)], [page('p3C', 7)]);
+    expect(stats.get(7)?.rmseFt).toBe(12);
+  });
+});
--- a/app/src/iiif/compare.ts
+++ b/app/src/iiif/compare.ts
@@ -60,7 +60,11 @@ export function statsByItemIndex(
   const byKey = new Map(rows.map((row) => [row.genPageKey, row]));
   const stats = new Map<number, PageCompareStats>();
   for (const page of pages) {
-    const row = byKey.get(page.stem);
+    // Sidecar keys are normalized to lowercase (compareTxt.ts); page stems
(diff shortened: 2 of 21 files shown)


danvk:
I'm going to abandon this until I see clearer evidence that it's needed and helpful.

Issue #275, opened 2026-08-08
snap: candidates.jsonl never expires records, so a re-split page keeps its whole-page fit


danvk:
Found while A/B-testing the #272 splitter change on grand_rapids. The snap channel resurrected a whole-page fit for a page that had just been split, and the IIIF carried both it and the panel fit under the same label.

## What happens

`artifacts/osm_snap/candidates.jsonl` is an incremental cache, and nothing ever removes a record from it:

- `cmd_candidates` computes the current `targets`, merges the fresh records into `existing` (every record previously in the file), and writes the union back — [osm_snap_experiment.py:1122](mapsnap/osm_snap_experiment.py#L1122). A stem that is no longer a target keeps its old record forever.
- `cmd_select` then branches on `record.get("fit_state")` — the value **stored in the record**, not recomputed from disk — [osm_snap_experiment.py:2278](mapsnap/osm_snap_experiment.py#L2278) and [:2291](mapsnap/osm_snap_experiment.py#L2291).

So when a page's split changes, `page_fit_state` correctly returns `"split"` and `cmd_candidates` correctly drops it from `targets` ([:1050](mapsnap/osm_snap_experiment.py#L1050)) — but its stale `fit_state: "fitted"` record is still in the file, and select happily arbitrates/refines it.

## Evidence

grand_rapids p841, split 1 → 2 panels by #272. Both panels were freshly OCR'd and `p841__1.georef.json` was written, so `page_fit_state("p841")` == `"split"`. Yet `p841.txt` shows the snap pass running on the whole page:

```
mode: arbitrate   status: ok   fit_state: fitted
incumbent: ver=1.1692 name=15/18 rmse=13ft(vs truth)
outcome: refine: candidate #1 accepted
```

and it materialized `p841.georef-osm.json`. The IIIF then contained two items with the same label `... p841 [2]` — the parent with a full-page selector, the panel with its polygon:

```
.../04023_07_1953-0841/georef       polygon 0,0 6660,0 6660,8070 0,8070   (parent)
.../04023_07_1953-0841__1/georef    polygon 1016,3758 1011,2550 ...       (panel)
```

The compare then cross-matched all three truth panels to the parent, so the split looked like a no-op (604.3 → 602.2 ft, 383.9 → 383.9 ft). After purging the three stale split-parent records, the same inputs give the real answer: truth p841__2 fits from `p841__1` at 13.1 ft and truth p841__1 becomes honestly unplaced.

[...]

Issue #240, opened 2026-08-06
mapsnap fit is not idempotent: it reads its own stale sidecars and oscillates between two states


danvk:
## Summary

`mapsnap fit` is **not reproducible at HEAD**: two runs over byte-identical inputs, with the same code, produce different georeferences for roughly half the pages and scores that differ by ~2 points. The pre-`2026-07-30` code **is** reproducible on the same data, so this was introduced by something in `4152898..main`.

This matters beyond tidiness: it sets a floor of about **+/-1.0 points** under every A/B, regression check and bisect we run, and it silently invalidated part of the `2026-08-05` rerun analysis until it was measured.

## Evidence

Grand Rapids, tonight's frozen data (no re-OCR between runs), `artifacts/*/candidates.jsonl` cleared before each run, scored with the same `mapsnap score`:

| commit | run | score | pages differing vs previous run |
|---|---|---|---|
| `4152898` (pre-07-30) | baseline-r1 | 60.1% | - |
| `4152898` | baseline-r2 | 60.1% | **0 of 63** |
| `4152898` | baseline-r3 | 60.1% | **0 of 63** |
| `main` (`10cd60c`) | head-r1 | 63.1% | - |
| `main` | head-r2 | 65.0% | **29 of 61** |
| `main` | 2026-08-05 | 64.2% | - |

Three runs at the old commit are byte-identical. Two runs at HEAD disagree on 29 of 61 published pages, and the resulting sidecar counts differ too (`georef-osm` 29 vs 28, plain `georef` 61 vs 60). No page changed *which channel* won, so the same channel is emitting different GCPs run to run.

A second, independent sighting: **Fargo** scored 47.2% in the afternoon and 45.5% at night with `raw/p0.georef.json`, `raw/p0.regions.panels.json`, `adjacency.json`, all 112 `p*.streets.json` and all 60 plain `p*.georef.json` **verified byte-identical** between the two runs. Only `snap`'s output moved.

## Consequences

1. **Small deltas are unmeasurable.** In the `2026-08-05` rerun, brooklyn (-0.4), hudson (-0.8), chicago (+0.6) and kansas_city (+0.4) are all inside the noise band and cannot be called either way from a single run.
2. **Bisects need repeats.** A single measurement per commit can easily localize a "regression" that is an unlucky draw.
[...]

PR #349, opened 2026-08-24
Fit starts with cold snap caches (#342): warm saves 3 min/corpus, costs ~2 points of variance

### LINKED ISSUE #342: Snap cache temperature flips pages: cold rerun of unmodified main moves KC 81.6 -> 79.8
Found during the #341 corpus A/B. Kansas City on **unmodified main**, PYTHONHASHSEED=0, with `artifacts/osm_snap/candidates.jsonl` + selections cleared (tag `main-cold`), scores **79.83** vs the 2026-08-14 archive's **81.62** — reproduced exactly by two other branches under the same cold protocol, so the delta is cache temperature, not code:

```
                08-14(warm)  main-cold  ss-only  runnerup-corpus2
KC net             0.8160      0.7983    0.7983    0.7983
  p566__1             7.4       282.3     282.3     282.3
  p575__2            21.4       126.4     126.4     126.4
  p552__2           144.1        None      None      None
```

washington_dc shows the same class (p126__2: 4.1 → 436.7, net −0.3). All movers are split panels graded under changed parent/sibling placements.

Implications:
- The incremental snap cache is not just a speedup: prior-run search state (including name-hit hash-order effects the cache froze in) is an **input** to selection, worth ~2 net points on KC. The warm 08-14 numbers are one arbitrary resolution of pool-sensitive selections; a cold rerun is another.
- Every A/B must hold cache state constant across arms (the corpus run did — all arms cold — which is how this surfaced). Comparing any cold rerun against the warm 08-14 archives mixes this variance into the diff, in both directions (champaign/LA showed the favorable side: two disasters became honest abstentions).
- This is the energy-arbitration pool-robustness issue (nashville p6 class) measured at volume scale; #270's reconcile — deterministic joint selection over a full hypothesis set — is the structural fix.

danvk:
Closes #342

is the incremental snap cache worth its invalidation hazards?

**Measurement** (per-stage wall-clock, all 18 truth volumes, like-for-like 2-up runs): snap stage totals **4,958 s warm vs 5,113 s cold** — the cache saves ~3 minutes per corpus run, within noise (cold is faster on LA, columbia, nashville).

**Why it can't save more inside a fit**: `clear_derived_sidecars` wipes every georef sidecar, georef rewrites them all, and `candidates_record_fresh` keys on `georef_mtime` — every fitted page's record is stale by construction, every run. The only records that survive across fits are the mtime-less demoted/failure classes, which are precisely the #342 cache-temperature flips (KC p566__1: 7.4 ↔ 282 ft on cache state alone, ~2 net points) and the #275/KC-p551 stale-record re-adoption class. The cache's entire cross-run surface inside fit is its bug surface.

**Change**: `clear_derived_sidecars` also removes `artifacts/osm_snap/candidates.jsonl`, `selection_*.jsonl`, and `artifacts/street_solve/candidates.jsonl` — the #240 idempotency principle extended to cache records. Every fit is cold by definition; the files remain useful within a run and for standalone `mapsnap snap` iteration. Side benefit: A/B protocols no longer need manual cache snapshotting/clearing, which every experiment this month has had to do by hand.

### DIFF
--- a/mapsnap/experiments.py
+++ b/mapsnap/experiments.py
@@ -402,18 +402,20 @@ def archive_run(
     # poses this run actually weighed (#270). Without these the baseline is
     # unrecoverable once any post-run experiment re-matches: the files live at
     # fixed paths and are overwritten in place.
+    # Globbed, not enumerated: clear_derived_sidecars deletes every
+    # candidates/selection file at the next fit's start, so anything the
+    # archive misses is destroyed rather than merely stale -- and the
+    # selection variants have grown before (selection_volume, selection_argmax
+    # were never in the old hardcoded list).
     for channel in ("osm_snap", "street_solve"):
         channel_dir = artifacts_dir / channel
-        for name in (
-            "candidates.jsonl",
-            "selection_arbitrate.jsonl",
-            "selection_union.jsonl",
+        for source in (
+            *sorted(channel_dir.glob("candidates.jsonl")),
+            *sorted(channel_dir.glob("selection_*.jsonl")),
         ):
-            source = channel_dir / name
-            if source.exists():
-                target_dir = run_dir / channel
-                target_dir.mkdir(exist_ok=True)
-                shutil.copy2(source, target_dir / name)
+            target_dir = run_dir / channel
+            target_dir.mkdir(exist_ok=True)
+            shutil.copy2(source, target_dir / source.name)
     if iiif_path is not None and iiif_path.exists():
         shutil.copy2(iiif_path, run_dir / iiif_path.name)
     if compare_txt is not None and compare_txt.exists():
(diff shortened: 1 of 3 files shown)


danvk:
Design question from review — should candidates files live under `artifacts/<tag>/` like other sidecars? **They already do**: `archive_run` copies both channels' candidates/selection files into the tag dir at fit completion (#270 provenance), so post-this-PR the layout is exactly the sidecar pattern (channel dirs = per-run scratch, tag dir = durable record). Writing directly into the tag dir was considered and rejected: it threads the run id through every stage subprocess, breaks the debugger's fixed-path reads and tagless standalone `mapsnap snap`, and violates the manifest-last complete-or-absent rule for tag dirs.

One real gap surfaced: the archive list was hardcoded to three filenames while runs also write `selection_volume.jsonl`/`selection_argmax.jsonl` — harmless staleness before this PR, **destroyed-only-copy after it**. Second commit archives by glob so everything the clear deletes is archived first, by construction.

PR #371, opened 2026-09-01
Snap searches a fitted page's own scale rung

danvk:
Motivated by looking at Richmond p365 in the snap debugger from #370. This seems like a nice win. 

## The gap

For a fitted page, `build_page_context` seeds the search **center** from the incumbent RANSAC pose but builds the **scale ladder** from the volume median plus (rarely) a key-map-region family rung — the incumbent's own scale was never a prior. On the 2026-08-28 baseline, **103 fitted pages have an incumbent on a rung no prior offered; 91 of them are correct incumbents (≤ 100 ft)** — richmond's 100 ft/in 365–389 run, nashville's small sheets, columbia's sectionals, asheville p3x/p43–p71A, grand rapids 8xx, LA p1499\*. On all 91, every candidate was born at the wrong size (refinement's scale solve only nudges ±4%), snap beat the incumbent 0 times, and refinement — the mid-tier improver — was structurally unreachable. The region-area family rung is no substitute: on trustworthy incumbents it matches the real scale 41/559 times (7%).

## The change

`with_incumbent_scale`: for fitted pages, append the incumbent's scale (`affine_m_per_px`, source `incumbent`) to the ladder unless an existing prior is already within ~16% (`SCALE_PRIOR_DEDUPE_LOG`, now shared with the printed-note rung). The volume median keeps priority; pages at the volume's scale see no change at all. The existing challenge/refine/rung-flip gates still arbitrate, so the 8 wrong-incumbent cases (KC p556, grand rapids p828 …) lose nothing — their ladder already had the right rung.

## richmond p365 (the establishing case)

| | ladder | top candidate | decision |
|---|---|---|---|
| before | volume-median 0.203 | select 0.78, ver 0.17, **463 ft** | keep |
| after | + incumbent 0.395 | select 2.28, ver 1.14, **10.0 ft** | **refine** (37 ft → 10 ft) |

The truth row scores 2.31, so the search now reaches it (verdict flips from *search problem* to *agrees*). Note evaluate_pose's frame half-size derives from the ladder's max scale, so fixed-pose scores shift by ~0.03 between the two runs — noise, not signal.

## A/B on the four volumes holding 83 of the 91 pages

See the results comment below: **+4.6 richmond, +6.6 asheville, ≈0 nashville/columbia**, 24 pages improved, 5 refine-adoption regressions, disasters unchanged; the control arm reproduced the 2026-08-28 baseline exactly.

Tests: `affine_m_per_px` on the fixture scale; `with_incumbent_scale` adds a missing 2× rung and dedupes a near-identical one. pyright clean.

### DIFF
--- a/app/src/components/SnapPanel.test.tsx
+++ b/app/src/components/SnapPanel.test.tsx
@@ -103,6 +103,10 @@ describe('SnapPanel phase-2 records', () => {
     const html = render(record);
     expect(html).toContain('<td>truth</td>');
     expect(html).toContain('0.68'); // the truth row's select score
+    // θ and px/ft are derived from every row's affine, the incumbent included.
+    expect(html).toContain('rotation in snap');
+    expect((html.match(/0\.0°/g) ?? []).length).toBeGreaterThanOrEqual(4);
+    expect((html.match(/0\.35<\/td>/g) ?? []).length).toBeGreaterThanOrEqual(4);
     expect(html).toContain(
       'a pose 14713 ft from truth outscores the truth pose',
     );
--- a/app/src/components/SnapPanel.test.tsx
+++ b/app/src/components/SnapPanel.test.tsx
@@ -1,0 +1,132 @@
+import { describe, expect, it } from 'vitest';
+import { renderToStaticMarkup } from 'react-dom/server';
+import { SnapPanel } from './SnapPanel';
+import type { SnapRecord } from '../snap';
+
+const affine: [number, number, number][] = [
+  [1e-5, 0, -77.43],
+  [0, -1e-5, 37.55],
+];
+
+// Shaped like richmond p311 in the 2026-08-28 baseline: a fitted page whose
+// alias outscores the true-location candidate and the truth pose (#325).
+const record: SnapRecord = {
+  target: 'p311',
+  status: 'ok',
+  fit_state: 'fitted',
+  width: 1412,
+  height: 2037,
+  has_truth: true,
+  margin: 0.0464,
+  incumbent: {
+    world_affine: affine,
+    verification: -0.708,
+    name: { score: 0.1728, n_labels: 9, n_hits: 4 },
+    effective_gcps: 2,
+    rmse_ft: 210.3,
+  },
+  truth_pose: {
+    world_affine: affine,
+    verification: 0.1163,
+    inlier_frac: 0.3365,
+    ncc_fine: 0.3551,
+    chamfer_mean_m: 17.26,
+    name: { score: 0.4279, n_labels: 9, n_hits: 5 },
+    select_score: 0.6776,
+  },
+  candidates: [
+    {
+      world_affine: affine,
+      center: [-77.47, 37.57],
+      select_score: 1.0207,
+      verification: 0.8249,
+      rmse_ft: 14713.1,
+      theta_deg: -43.96,
+      scale_source: 'volume-median',
+    },
+    {
+      world_affine: affine,
+      center: [-77.43, 37.55],
+      select_score: 0.9743,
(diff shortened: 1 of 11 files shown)


danvk:
## A/B results: four volumes, 83 affected pages

Harness: two scratch copies per volume (small files copied with mtimes, images symlinked); **control** = the 2026-08-28 caches as-is, **treat** = the 83 pages whose ladder gains an incumbent rung recomputed on this branch; both arms then ran `fit`'s post-georef stages identically (snap → street-solve → reconcile --publish → iiif → compare → score). **The control arm reproduced the baseline exactly** — 0 pages moved >5 ft in all four volumes — so every treatment delta below is signal.

| volume | control score | treat score | Δ | ≤25 ft share | pages moved >5 ft |
|---|---|---|---|---|---|
| richmond | 78.6 | **83.2** | **+4.6** | 81.1% → 85.7% | 11 improved, 0 worse |
| asheville | 39.0 | **45.6** | **+6.6** | 43.2% → 49.7% | 8 improved, 1 worse |
| nashville | 69.7 | 71.2 | +1.5 (see caveat) | 69.9% → 71.4% | 2 improved, 2 worse |
| columbia | 74.7 | 74.8 | +0.1 | 76.2% → 76.2% | 3 improved, 2 worse |

Disaster share unchanged in every volume. Band transitions: richmond 3 fair→good + 1 poor→good; asheville 4 fair→good + 2 poor→good + 1 poor→fair, against 1 fair→poor; columbia 1 fair→good against 1 good→fair; nashville 1 poor→good (p24, caveat below).

**Wins are the predicted ones** — refinement finally reaching correct-but-off-ladder incumbents: richmond p365 37→10, p366 21→8, p367 45→7, p374 36→8, p377 64→25, p389 19→8; asheville p43 75→29, p58 77→15, p68 82→19, p60 30→11, p57 34→14, p56 36→21. Every one is a `refine` adoption at the `incumbent` rung.

**The five regressions are all `refine` adoptions too** — a challenger that out-verified the incumbent (Δver +0.07 to +0.64) but landed a little farther from truth: nashville p49 4.7→21.9 (ver 0.56→0.63, the thinnest margin) and p40 11.4→17.9; columbia p53 21.3→27.5 (crossing the 25 ft bar) and p46 33→40; asheville p66 48→61. In each, the truth pose's own select is *below* the challenger's, so the page's evidence genuinely prefers the slightly-off pose — refinement's known winner's-curse cost, now payable on pages that were previously immune only because no candidate could agree with them. Nothing in this PR makes that worse elsewhere; it's the price of admission to the 24 wins.

**Caveat on nashville p24 (171→15 ft):** its per-page verdict is `keep` and its top candidate is the *same* volume-median pose in both arms (14–15 ft, ver ~0.70); the move came through reconcile's joint solve flipping a near-tie (control published `georef` at −1.372, treat `snap:0` at −1.302 — evaluation drift or the refined neighbours shifting the pairwise terms). I'm not crediting it to the rung; without it nashville is ≈0.

**Truth-row census on the 83 pages after the change** (phase-2 records): richmond 15/16 *agrees*; nashville 19/22; columbia 12/17; asheville 13/28 — leaving **15 residual search problems** (truth still outscores every candidate: 9 asheville, 4 columbia, 1 each richmond/nashville) and 9 alias cases. The ladder was the dominant blocker but not the only one; those 15 are the next thing the truth row points at.

The remaining 8 affected pages live in other volumes (grand rapids 8xx, hudson p3/p73/p74, LA p1407/p1499\*, NO-1951 p531) and will pick this up on the next baseline. Reminder: cached records don't recompute on code changes, so a baseline needs a fresh snap (or `--recompute`) to see this.

Issue #144, opened 2026-07-21
Consider "inappropriate" orientations for CRAFT


danvk:
See #143 

For Brooklyn 1939 v2 p19 (the repo example image), only a 0° and 270° rotation produce a full box for "TILLARY," whereas 90° (the natural orientation) cuts it in half:

<img width="499" height="227" alt="Image" src="https://github.com/user-attachments/assets/9b6bde3f-7add-4c14-b7a3-923228fa764d" />

It's unclear how widespread this phenomenon is, but if a box doesn't produce a confident read, it might be worth checking whether the other orientations produce an overlapping but differently-sized box and using those instead.

danvk:
Richmond p365 is another example of this. The "appropriate" orientation gloms the 373 adjacency stamp together with HAMMOND, but the other orientations do not.

<img width="865" height="218" alt="Image" src="https://github.com/user-attachments/assets/10b4a7bd-865c-4192-8225-0cf59f10da41" />

danvk:
Also Detroit p85. The correct orientation has "S. TENNESSEE", which fails to match TENNESSEE at high confidence, but the inappropriate orientations segment differently:

<img width="462" height="219" alt="Image" src="https://github.com/user-attachments/assets/69a2ff04-faf7-4e64-b1bc-7a094afac8cd" />

danvk:
## Scope and risk assessment (prototype on the three motivating pages + corpus census)

### Mechanism

The recognizer already reads every angle's CRAFT boxes, but rejects any box that is taller than wide *in that angle's frame* (`_recognize_pass`). So a footprint that only the "wrong" orientations segmented well is never read in the orientation where the text is upright — that is the gap. The fix is a **cross-orientation transfer**: take the other angle's footprint (horizontal box or free polygon), rotate it into the natural angle's frame, and recognize it there. No new detection, ~one extra recognition per trigger.

### Prototype: 3 of 3 motivating pages rescue, at 0.9–1.0

| page | natural-orientation read today | other-orientation footprint | transferred read (natural frame) |
|---|---|---|---|
| detroit p85 TENNESSEE | 92×24 box `S. TENNESSEE` → `TENNESSEE@0.162` | 270° box 71×24 (drops the "S.") | **`TENNESSEE@0.899`** |
| richmond p365 HAMMOND | 176×64 box glomming the "373" stamp → `13TH@0.002` | 90° split box 97×31 | **`HAMMOND@1.000`** |
[...]

PR #372, opened 2026-09-01
OCR re-reads failed large boxes via cross-orientation transfer

danvk:
Implements #144. CRAFT segments the same text differently at 0°/90°/270°, and the recognizer only ever reads each orientation's own boxes — so when the natural orientation cuts a word in half (brooklyn v2 p19 TILLARY), gloms it with an adjacency stamp (richmond p365 "373 HAMMOND"), or attaches a poisoning prefix (detroit p85 "S. TENNESSEE"), the label is lost even though another orientation segmented it correctly.

## The transfer pass

After the normal primary+fallback vocab passes, any **large wide box** (≥80 px, aspect ≥2 in its own frame) whose best read is **below 0.3**:
- has the *other* orientations' clean **split/merge** footprints re-read in its frame, and
- has its own footprint re-read in the other frames,

through the same primary+fallback vocab structure. Horizontal boxes transfer as rotated bboxes; free polygons as **reading-ordered quads** (EasyOCR's `four_point_transform` takes the first edge as the crop width, so rotated quads must be re-ordered or the crop comes out mirrored — TILLARY read `'E'` until this). Quads keep the true label orientation in the output polygon (`dir_pix` 1.29 rad for the diagonal TILLARY), which georef needs.

Transferred reads are flagged `transferred` / `transfer_kind` (footprint|split|merge) / `transfer_source_angle`, floored at 0.2 confidence, and deduped: same-text overlapping pairs keep the higher confidence, with near-ties (<0.02) preferring the existing read so a transferred twin never masquerades as new. `--no-transfer` disables the pass (A/B control, debugging).

## Evidence so far (10-page before/after sample, reviewed)

Pickups: TILLARY@1.00, HAMMOND@1.00, TENNESSEE@1.00, THALIA@0.64 (possibly fit-critical for NO-1896 p119), BRIDGE@0.95, ST@1.00. Residual noise: pipe annotations (`S W PIPE`/`EW PIPE`), `PRESS`×2, and city/title words via the **fallback vocab** (RICHMOND, CHICAGO, STATE) — a "no transferred reads from the fallback pass" gate would remove those, deliberately **not** included yet pending evidence of real damage (per review).

Unit tests: reading_order, trigger/footprint construction, merge/dedupe/floor semantics. Single-page CLI check: p365 with transfer gains exactly HAMMOND@1.00 + RICHMOND@0.82 flagged; `--no-transfer` output is unchanged production behavior.

## A/B status — two volumes, no win

| volume | control | transfer | delta |
|---|---|---|---|
| new_orleans_la_1896_vol_2 | 78.7 | 76.4 | −2.3 |
| richmond_va_1925_vol_3 | 83.2 | 82.0 | −1.2 |

Neither delta is caused by a bad read: NO-1896's two regressions are on pages whose reads are **byte-identical** between arms (snap coupled through a 0.2% shift in the volume-calibrated search radius), and Richmond's entire −1.2 is **one page moving 0.9 ft** (p377, 24.6 → 25.5) across the band line via RANSAC-consensus jitter.

The motivating rescue works and doesn't help: richmond p365 gains `HAMMOND`@1.00 as a 16th matched label (+6 intersections) and moves **9.9 → 10.7 ft** — it already had 15 inliers. Across ~230 transferred reads on two volumes, no page's fit materially improved.

**Not merging as default-on.** The promising variant is to gate transfer on **label-starved pages** (<5 matched labels or <2 intersections), where a recovered street can actually change the outcome — see the Richmond comment for the full argument.

### DIFF
--- a/mapsnap/ctc_vocab_decode.py
+++ b/mapsnap/ctc_vocab_decode.py
@@ -24,6 +24,7 @@ from __future__ import annotations
 import re
 from dataclasses import dataclass, field
 from functools import lru_cache
+from typing import Any
 
 import numpy as np
 
@@ -204,6 +205,14 @@ def generate_vocab_strings(normalized_streets: set[str]) -> list[str]:
 class TrieNode:
     children: dict[str, TrieNode] = field(default_factory=dict)
     is_end: bool = False
+
+
+# The TRUE originals of the functions patch_easyocr_reader replaces, stashed on
+# first patch. Re-patching happens on every vocabulary switch (several times per
+# page); wrapping the previous wrapper instead grew the call stack by one frame
+# per patch and crashed long OCR runs with RecursionError inside torch's LSTM
+# (found by the #144 A/B, where the transfer pass doubles the patch rate).
+_PATCH_ORIGINALS: dict[str, Any] = {}
 
 
 def build_trie(strings: list[str]) -> TrieNode:
@@ -429,10 +438,9 @@ def patch_easyocr_reader(
     # wrapper and the growing call stack eventually hits Python's recursion
     # limit mid-inference (observed at ~1,000 patches: RecursionError inside
     # torch's LSTM on the #144 A/B, where the transfer pass doubles the rate).
-    original_predict = getattr(_recog, "_mapsnap_original_recognizer_predict", None)
-    if original_predict is None:
-        original_predict = _recog.recognizer_predict
-        _recog._mapsnap_original_recognizer_predict = original_predict
+    original_predict = _PATCH_ORIGINALS.setdefault(
+        "recognizer_predict", _recog.recognizer_predict
+    )
 
     def _patched_recognizer_predict(
         model,
@@ -525,10 +533,9 @@ def patch_easyocr_reader(
     import easyocr.utils as _eu
     from easyocr.recognition import get_text as _get_text
 
-    original_recognize = getattr(easyocr.Reader, "_mapsnap_original_recognize", None)
-    if original_recognize is None:
-        original_recognize = easyocr.Reader.recognize
-        easyocr.Reader._mapsnap_original_recognize = original_recognize
+    original_recognize = _PATCH_ORIGINALS.setdefault(
+        "recognize", easyocr.Reader.recognize
+    )
 
     def _patched_recognize(
         self,
--- a/mapsnap/ctc_vocab_decode.py
+++ b/mapsnap/ctc_vocab_decode.py
@@ -424,7 +424,15 @@ def patch_easyocr_reader(
     trie_root = build_trie(vocab_strings)
     char_list: list[str] = reader.converter.character
 
-    original_predict = _recog.recognizer_predict
+    # Re-patching happens on every vocabulary switch (several times per page).
+    # Wrap the TRUE original, stashed once, or each call would nest another
+    # wrapper and the growing call stack eventually hits Python's recursion
+    # limit mid-inference (observed at ~1,000 patches: RecursionError inside
+    # torch's LSTM on the #144 A/B, where the transfer pass doubles the rate).
+    original_predict = getattr(_recog, "_mapsnap_original_recognizer_predict", None)
+    if original_predict is None:
+        original_predict = _recog.recognizer_predict
+        _recog._mapsnap_original_recognizer_predict = original_predict
 
     def _patched_recognizer_predict(
         model,
@@ -517,7 +525,10 @@ def patch_easyocr_reader(
     import easyocr.utils as _eu
     from easyocr.recognition import get_text as _get_text
 
-    original_recognize = easyocr.Reader.recognize
+    original_recognize = getattr(easyocr.Reader, "_mapsnap_original_recognize", None)
+    if original_recognize is None:
+        original_recognize = easyocr.Reader.recognize
+        easyocr.Reader._mapsnap_original_recognize = original_recognize
 
     def _patched_recognize(
         self,
--- a/mapsnap/ctc_vocab_decode_test.py
+++ b/mapsnap/ctc_vocab_decode_test.py
@@ -1,5 +1,7 @@
 """Tests for ctc_vocab_decode helpers."""
 
+from typing import ClassVar
+
 import numpy as np
 
 from mapsnap.ctc_vocab_decode import (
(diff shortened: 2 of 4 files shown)


danvk:
## A/B results — new_orleans_la_1896_vol_2 (fresh OCR both arms)

**Verdict: no volume-level win, and the −2.3 that shows up is not read damage — it is second-order snap coupling on pages whose reads never changed.** Detail below; the honest headline is that this volume could only ever measure downside, because none of the three motivating pages lives in it.

| arm | score | ≤25 ft | 25–50 | 50–200 | ≥200 |
|---|---|---|---|---|---|
| control (`--no-transfer`) | **78.7%** | 81.1% | 5.6% | 7.6% | 2.4% |
| transfer, floor 0.2 | 76.4% | 78.8% | 6.7% | 8.8% | 2.4% |
| transfer, floor 0.5 | 76.4% | 78.8% | 6.7% | 8.8% | 2.4% |

The control reproduces the 2026-08-28 baseline exactly (78.7), so the harness is sound.

### What the feature actually did to this volume

74 transferred reads survived on 44 of 114 pages (floor 0.5). Corpus-wide, exactly **one** confident control read was replaced (p119 `THALIA` 0.45 → 0.64, same box within 3 px), and **zero** pages improved. Of 114 published poses, only 3 differ from control by more than 2 m:

| page | ctl → treat | transferred reads on that page? |
|---|---|---|
| p101s__1 | 17 m (same band) | yes |
| p128 | 3 m (same band) | **no** |
| p162 | 6 m, 15.7 → 55.5 ft (good → poor) | **no** |

### The two regressions are both read-independent

- **p162** — its 175 reads are **byte-identical** between arms (verified full-record compare), and its snap record is fresh (`georef_mtime` matches the post-rerun `georef.json`, so nothing was cached). Yet control selects the true pose (`sel 1.073 / ver 0.438 / rmse 15.7`, verdict `refine`) and treatment selects an alias (`sel 1.293 / ver 0.696 / rmse 476.7`, verdict `keep`, published raw georef at 55.5 ft). The only visible input difference is the snap **search radius: 164.1 m → 163.8 m** — a 0.2% shift, because `calibrated_radius_m` is computed from the *whole volume's* keymap residuals and other pages' fits moved fractionally. A 0.2% radius change resamples the pose grid, finds an alias 477 ft away that verifies **0.696 vs the true pose's 0.431**, and selection prefers it despite 3 name hits vs 6. That is a snap-selection fragility worth its own issue; it is not evidence about OCR transfer.
- **p191** — 21.5 → 25.5 ft (good → fair), also with identical reads; neighbor-coupled refinement drift through the `ransac-neighbor` prior. 4 ft of noise that happens to straddle the 25 ft line.

Together those two pages are exactly the 2.3 points of ≤25 ft that separate the arms (disasters identical at 2.4% in both).

### Floor 0.2 → 0.5

Raised after the first pass, since every reviewed pickup is ≥0.64 and the sub-floor reads were where the junk lived. It removed all sub-0.5 transfers from the 15 affected pages (p162's `ORLEANS`@0.49 among them) — and the volume score did not move, because p162's flip turned out not to depend on that read at all. Keeping the higher floor anyway: it costs nothing here and drops the weakest class of transfer.

### On the fallback-vocab gate

25 of the 74 confident transfers came through the fallback vocab (RICHMOND, CHICAGO, STATE class). **None of them caused an observed regression in this run** — the two regressions have no transferred reads on the page. So there is still no evidence that the "no transferred reads via fallback" rule is needed, which was the condition for holding it back.

### Recommendation

This volume was a poor upside test: TILLARY is brooklyn v2 p19, HAMMOND is richmond v3 p365, TENNESSEE is detroit v11 p85, and NO-1896's only pickup (`THALIA`) was **already read and matched at 0.45 in control** — p119 has 5 street labels but only 1 usable intersection either way, so it stays on `snap:0` at ~702 ft in both arms. The meaningful next A/B is richmond_va_1925_vol_3 or brooklyn_ny_1939_vol_2, where a motivating rescue actually lives. Holding this PR until then.

danvk:
## A/B #2 — richmond_va_1925_vol_3 (the volume with a motivating page)

| arm | score | ≤25 ft | 25–50 | 50–200 | ≥200 |
|---|---|---|---|---|---|
| control (`--no-transfer`) | **83.2%** | 85.7% | 6.6% | 3.2% | 2.5% |
| transfer (floor 0.5) | 82.0% | 84.5% | 7.8% | 3.2% | 2.5% |

(Control is 83.2 rather than the 08-28 baseline's 78.6 because #371 merged this afternoon and is an ancestor of this branch — an independent reproduction of that PR's +4.6 on Richmond, on a fresh OCR pass. Both arms carry it, so the comparison is clean.)

### The entire −1.2 is one page moving 0.9 ft

**Zero pages moved more than 5 ft.** One band change: p377, **24.6 → 25.5 ft**. That page carries 0.2685 land weight, which is the whole 1.2-point delta. Only 3 of 95 published poses differ by more than 2 m (p308, p347, p370__1 — all 2–3 m, all within band).

p377's mechanism is fit jitter, not evidence: the two transferred reads are `RICHMOND`@0.97 (**fallback vocab**) and `ONE INCH`@0.87 (already `ignore: true`), appended at the end of the read list; neither is in the page's matched-label set. What changed is the RANSAC consensus — `EDGEWOOD`, read **identically in both arms at 0.956**, joins the model, and the fit shifts 0.9 ft across the band line.

### The motivating rescue works, and doesn't matter

p365 is the `HAMMOND` page from #144. Transfer does exactly what it was designed to do:

| | control | transfer |
|---|---|---|
| matched street labels | 15 (15 inliers) | **16** (16 inliers, +`HAMMOND`) |
| intersections | 33 | **39** |
| rmse | **9.9 ft** | 10.7 ft |

`HAMMOND`@1.00 is recovered, matched, and becomes an inlier — and the page moves 0.8 ft the wrong way, because at 15 inliers and 33 intersections it was never short of evidence. **That is the finding**: the reads #144 recovers are real, but they land on pages that already have plenty. A 16th label cannot rescue a page that is already at 9.9 ft.

### Transfer census

159 transferred reads on 76 of 96 pages. **87 of them are boilerplate**: `RICHMOND` ×44 (the volume title) and `SCALE` ×43 (scale-bar text). Real street pickups are a handful — `WILLIAMS`, `JACKSON`, `BAKER`, `BROOKLAND`, `CRAWFORD`, `CUSTER`, `3RD`. 47 of 159 came through the fallback vocab.

### Evidence on the fallback gate

You asked to hold the "no transferred reads via fallback" rule until there was evidence that its absence does real damage. Across both volumes, exactly one transferred read is implicated in any score movement — p377's `RICHMOND`@0.97 — and it **is** a fallback-vocab read. That is real but weak evidence (0.9 ft of jitter, not a broken fit), so it argues for the gate without being the damage case you were waiting for.

## Where this leaves #144

Two volumes, ~230 transferred reads, and **not one page's fit materially improved**. Every score movement observed is either sub-1-ft jitter that happens to straddle a band line (richmond p377, NO-1896 p191) or a chaotic snap flip on a page whose reads never changed (NO-1896 p162). Net: −2.3 and −1.2, both noise-shaped, neither caused by a bad read.

I don't think this should merge as a default-on change on this evidence. Three options, in the order I'd rank them:

1. **Gate transfer on label-starved pages** — run it only where a new street can change the outcome (say <5 matched labels or <2 intersections), not on pages already carrying 15 inliers. NO-1896 p119 (5 labels, 1 intersection, unfittable) is that case; richmond p365 (15 labels) is not. This turns a volume-wide perturbation into a targeted rescue and removes essentially all of the observed jitter.
2. Keep it behind the flag, **off by default**, as a debugging/rescue tool.
3. Add the fallback gate and re-measure — cheaper, but it doesn't address the core finding that the pickups land on healthy pages.

danvk:
This seems not worth it, at least for now.

PR #376, opened 2026-09-01
Charge snap candidates for street names that match nothing

### LINKED ISSUE #375: Snap discards the negative half of street-name evidence
Snap's shape evidence (P(road) chamfer + NCC) aliases badly in regular street grids, and we guard against it with gates and thresholds. But we already extract the perfect anti-alias signal — the street *names* — and then throw away its negative half.

## The term today

[`name_alignment`](mapsnap/osm_snap.py) projects each unambiguous OCR label through a candidate pose, snaps it to its **own** street's polyline, and counts a hit within 60 m and 25°:

```
score = Σ exp(-dist / 25 m) / (n_labels + 2)
```

It feeds [`selection_score`](mapsnap/osm_snap.py) at `W_NAME = 1.0` — the largest bonus weight — and `evaluate_pose` computes it, so arbitration and `reconcile`'s unary energy see it too. Labels also steer the rotation-prior ladder. So names are *used*; this is not a P(road)-only system.

**But the term is floored at zero, by design.** From the docstring: *"the +2 in the denominator keeps one lucky label from dominating while a renamed street (no match anywhere) scores 0, never negative."* A pose with nine legible street labels, none of which lands anywhere near a street of that name, scores exactly the same as a page with no labels at all: **0.0**.

## richmond p311: a 14,713 ft disaster that nine labels contradict

| candidate | grid rmse | verification | name | select |
|---|---|---|---|---|
| C0 (published) | **14,713 ft** | 0.825 | **0.000 — 0 of 9 labels hit** | **1.021** |
| C2 | **18.8 ft** | 0.229 | 0.410 — 5 of 9 hit | 0.974 |

C2's hits are EAST PRESTON, NORTH SEVENTH, EAST BATES, NORTH FOURTH, NORTH FIFTH — the correct pose was **sitting at rank 3** and lost by **0.046**. The alias banks a strong grid-shape score; the truth can only add a bonus on top of weak P(road) evidence. Nine labels saying "this is not the north side" cost the alias nothing.

## How much signal is on the floor

Over **12,213 cached snap candidates** with ≥3 eligible labels (every volume's `artifacts/osm_snap/candidates.jsonl`):

| pose quality | zero name hits | median hit fraction |
|---|---|---|
| accurate (≤50 ft) | **0.8%** (14 / 1749) | **0.91** |
| wrong (≥500 ft) | **43.3%** (1568 / 3622) | 0.14 |

"Zero hits despite ≥3 eligible labels" is roughly a **50:1 likelihood ratio** against the pose being correct, and it is currently worth zero points.

## Proposal — make the term signed

Charge for eligible labels that match nothing, in the same denominator:

```
evidence = (Σ exp(-dist / τ) − W_NAME_MISS × (n_labels − n_hits)) / (n_labels + 2)
```

with `W_NAME_MISS = 0.5` and the penalty inert below `NAME_MISS_MIN_LABELS = 3` eligible labels (a page with one or two labels says nothing either way). Reward half is unchanged, so a page with no labels still scores 0 and nothing regresses on label-less pages.

## Offline re-ranking on cached candidates

The code comment invites exactly this: *"every term is logged per candidate so a re-ranking experiment can refit them on cached candidates."* Re-ranking all cached candidates with the signed term, no re-running of snap:

- top candidate changes on **29 of 1,772** pages: **24 materially better (≥20% closer), 0 materially worse (≥25% farther)**
- **16 disaster→good, 0 good→disaster**

| volume | page | before | after |
|---|---|---|---|
| richmond_va_1925_vol_3 | p311 | 14,713 ft | **18.8 ft** |
| richmond_va_1925_vol_3 | p323 | 2,373 ft | **22.7 ft** |
| brooklyn_ny_1939_vol_1 | p8 | 519 ft | **7.6 ft** |
| columbia_sc_1956_vol_1 | p21 | 511 ft | **19.2 ft** |
| new_orleans_la_1896_vol_2 | p190 | 487 ft | **13.3 ft** |
| washington_dc_1916_vol_2 | p221 | 449 ft | **9.1 ft** |
| new_orleans_la_1896_vol_2 | p136 | 436 ft | **16.4 ft** |
| washington_dc_1916_vol_2 | p207 | 426 ft | **25.6 ft** |
| columbus_oh_1951_vol_3 | p244 | 409 ft | **13.7 ft** |
| fargo_nd_1958 | p56 | 376 ft | **12.8 ft** |
| new_orleans_la_1896_vol_2 | p127 | 377 ft | **13.2 ft** |
| kansas_city_mo_1951_vol_4 | p537 / p538 | 335 / 337 ft | **33 / 37 ft** |

## Known limits and risks

- **Gate recalibration.** Every score with misses shifts down, so `PRODUCTION_GATE_SCORE = 1.35` and friends are no longer calibrated. Against today's constants, 8 of the 16 rescues publish and 8 fall below the gate and abstain — still a disaster avoided, but coverage lost. The A/B must report both, and gate re-derivation is part of the work.
- **Accurate poses that legitimately match nothing**: 14 of 1,749 (9 los_angeles_ca_1949_vol_14, 4 miami_fl_1950_vol_1, 1 fargo_nd_1958) — presumably renamed streets or restricted-vocab misses. This is exactly the case the zero floor was protecting; `NAME_MISS_MIN_LABELS` and `W_NAME_MISS` are the levers, and the sim shows none of them lost their crown.
- **Not a cure-all for aliasing.** `name_alignment` measures distance to the *polyline*, so sliding along a long street is free. NO-1896 p162's alias (477 ft) still hits 3 of 8 labels; the penalty narrows its lead from 0.229 to 0.079 but does not flip it. Cross-street contradiction ("this label sits on a street with a *different* name") is a stronger signal still, and is not proposed here.
- **Winner's curse**: the cached candidate sets were produced by a search steered by the current score, so a truth pose only appears if the search happened to reach it (p311's did, via three keymap search centers). Re-ranking cannot rescue a pose the search never generated.
- Simulation distances are grid-rmse, the cheap diagnostic that diverges from region-graded rmse on split panels. Gate decisions must be judged on real `mapsnap compare` output.
- `W_NAME_MISS = 0.5` / `MIN_LABELS = 3` are first guesses with no sweep.

## Plan

1. Shared `name_evidence()` helper; `W_NAME_MISS`, `NAME_MISS_MIN_LABELS`; signed value recorded alongside the raw score in `candidates.jsonl` so nothing is lost for later refits.
2. Wire into `SnapCandidate.select_score`, `rank_pose`, and `reconcile`'s unary evidence term (it already imports `W_NAME`).
3. Unit tests: zero-hit penalty, exempt below the label floor, reward half unchanged, no-label pages still 0.
4. Volume A/B against a base-commit worktree (richmond first — it owns two of the rescues), then a weight sweep and gate re-derivation.

danvk:
Prototype for #375: snap's street-name term becomes signed evidence instead of a reward-only bonus.

## The change

`name_alignment` scores `Σ exp(-d/25 m) / (n_labels + 2)` over labels that land on their own street, floored at zero — so a pose whose labels match **nothing** scores identically to a page with no labels at all. `name_evidence` undoes the denominator, charges `W_NAME_MISS = 0.5` per eligible label that found no match, and restores it:

```
evidence = (score × (n_labels + 2) − W_NAME_MISS × (n_labels − n_hits)) / (n_labels + 2)
```

Below `NAME_MISS_MIN_LABELS = 3` eligible labels the old reward-only value stands (one unmatched label is noise, not contradiction), and the reward half is untouched, so nothing changes for label-less pages.

Wired into `SnapCandidate.select_score`, `rank_pose`, and `reconcile`'s unary evidence term (it already imported `W_NAME`). Candidate records now carry `name.evidence` beside `name.score`; `name_evidence_of` recomputes it for records written before this change, so cached `candidates.jsonl` from earlier runs still ranks correctly.

## Motivating case — richmond p311

| candidate | grid rmse | verification | name | select |
|---|---|---|---|---|
| C0 (published) | **14,713 ft** | 0.825 | 0.000 — **0 of 9 labels hit** | **1.021** |
| C2 | **18.8 ft** | 0.229 | 0.410 — 5 of 9 hit | 0.974 |

The correct pose sat at rank 3 and lost by 0.046. Nine labels saying "this is not the north side" cost the alias nothing. A unit test pins this exact comparison.

## Offline evidence — `scripts/name_evidence_sim.py`

Re-ranks every cached snap candidate without re-running snap (the code comment invites exactly this: *"every term is logged per candidate so a re-ranking experiment can refit them on cached candidates"*). Over 14,093 candidates in 24 volumes:

```
candidates with >=3 eligible labels:
  good:   1749 candidates, zero hits    14 (  0.8%)
  bad :   3622 candidates, zero hits  1568 ( 43.3%)
  good hit fraction: median 0.91   bad: 0.14

miss cost 0.5, label floor 3: 29 pages change winner
  materially better (>=20% closer): 24
  materially worse  (>=25% farther): 0
  disaster -> good: 16   good -> disaster: 0
```

Rescues include p311 14,713→18.8 ft, richmond p323 2,373→22.7, brooklyn p8 519→7.6, DC p221 449→9.1, columbus p244 409→13.7, fargo p56 376→12.8.

### Weight sweep

| miss cost | flips | better | worse | disaster→good | good→disaster |
|---|---|---|---|---|---|
| 0.25 | 17 | 13 | 0 | 8 | 0 |
| **0.5** | **29** | **24** | **0** | **16** | **0** |
| 0.75 | 41 | 30 | 4 | 21 | **2** |
| 1.0 | 50 | 36 | 6 | 23 | **3** |

There is a clean cliff: 0.5 is the largest cost with zero observed regressions, which is why it is the default. The label floor is the weaker lever (3 vs 2 costs 5 flips and no rescues; 4 costs two rescues).

## Not yet done

- **Volume A/B.** The numbers above are offline re-ranking on grid rmse — the cheap diagnostic that diverges from region-graded rmse on split panels. A real `fit` A/B (control from a base-commit worktree, both arms with snap caches cleared) is the next step; richmond first, since it owns two rescues.
- **Gate recalibration.** Scores with misses shift down, so `PRODUCTION_GATE_SCORE = 1.35` and friends are no longer calibrated. Judged against today's constants, 8 of the 16 rescues clear the gate and 8 fall below it and abstain — a disaster avoided but coverage lost. The A/B will report both, and re-deriving the gates is a follow-up.
- **Aliasing is only half-addressed.** `name_alignment` measures distance to the polyline, so sliding *along* a street is free: NO-1896 p162's 477 ft alias still hits 3 of 8 labels and the penalty narrows its lead from 0.229 to 0.079 without flipping it. Cross-street contradiction (a label sitting on a street with a *different* name) is a stronger signal and is not attempted here.

1,086 tests pass; ruff and pyright clean.

Fixes #375

### DIFF
--- a/mapsnap/osm_snap.py
+++ b/mapsnap/osm_snap.py
@@ -20,6 +20,7 @@ objects, and evaluates against truth; nothing in this module reads truth data.
 
 import dataclasses
 import math
+import os
 from dataclasses import dataclass, field
 
 import cv2
@@ -69,7 +70,12 @@ CALIBRATED_RADIUS_MARGIN_M = 100.0
 CALIBRATED_RADIUS_MIN_M = 150.0
 
 # select_score weights (hand-tuned; every term is logged per candidate so a
-# re-ranking experiment can refit them on cached candidates).
+# re-ranking experiment can refit them on cached candidates). The two
+# name-miss knobs read the environment (MAPSNAP_NAME_MISS,
+# MAPSNAP_NAME_MISS_MIN_LABELS) so an A/B or a sweep is ONE build with the arm
+# chosen at run time: `mapsnap fit` spawns every stage as a subprocess, and
+# the environment crosses that boundary where a CLI flag would have to be
+# threaded through each stage.
 W_NAME = 1.0
 W_CONTAIN = 0.3
 W_PRIOR = 0.1
@@ -88,10 +94,22 @@ W_PRIOR = 0.1
 # select score and stopped clearing PRODUCTION_GATE_SCORE, and brooklyn p9's
 # correct pose (1 of 5) was demoted under an alias that hits 4 of 5 because
 # sliding along a long avenue keeps a label near its own street.
-W_NAME_MISS = 0.5
+W_NAME_MISS = float(os.environ.get("MAPSNAP_NAME_MISS", "0.5"))
 # Below this many eligible labels a page says nothing either way, so the
-# penalty stays inert (one unmatched label is noise, not contradiction).
-NAME_MISS_MIN_LABELS = 3
+# penalty stays inert (a couple of unmatched labels are noise, not
+# contradiction). Measured on 13,326 fresh candidates across the 20 truth
+# volumes, raising the floor sharpens the signal because thin-evidence pages
+# are where an accurate pose legitimately matches nothing:
+#
+#   floor   accurate poses w/ zero hits   wrong poses w/ zero hits   ratio
+#     3                0.8%                        43.6%              51:1
+#     5                0.4%                        39.4%             105:1
+#     6                0.2%                        36.3%             199:1
+#
+# Every page the corpus A/B damaged carried exactly 3 eligible labels
+# (detroit p93, brooklyn p13, miami p74 -- a correct 32.8 ft pose matching
+# 0 of 3); both rescues carry 9 and 10 (richmond p311, p323).
+NAME_MISS_MIN_LABELS = int(os.environ.get("MAPSNAP_NAME_MISS_MIN_LABELS", "5"))
 
 # The recipe validated by the issue-#128 exploration: generous overlap window
 # (the page may sit entirely inside the OSM frame, unlike an edge join) and a
--- a/mapsnap/osm_snap.py
+++ b/mapsnap/osm_snap.py
@@ -74,6 +74,18 @@ W_NAME = 1.0
 W_CONTAIN = 0.3
 W_PRIOR = 0.1
 
+# Cost per eligible label that matches nothing at a pose (issue #375). The
+# reward half of name_alignment is floored at zero, so a pose whose labels all
+# land on the wrong ground scores the same as a page with no labels at all --
+# and a grid alias with strong P(road) shape evidence outranks the truth
+# (richmond p311: 0 of 9 labels hit, published 14,713 ft off, beating the
+# 18.8 ft pose by 0.046). Corpus-wide, zero hits with >=3 eligible labels
+# occurs in 0.8% of accurate poses and 43.3% of >=500 ft poses.
+W_NAME_MISS = 0.5
+# Below this many eligible labels a page says nothing either way, so the
+# penalty stays inert (one unmatched label is noise, not contradiction).
+NAME_MISS_MIN_LABELS = 3
+
 # The recipe validated by the issue-#128 exploration: generous overlap window
 # (the page may sit entirely inside the OSM frame, unlike an edge join) and a
 # deeper top-K since ranking happens downstream.
@@ -107,12 +119,53 @@ class ScalePrior:
 
 @dataclass
 class NameAlignment:
-    """OCR street-name agreement with a candidate pose (a boost, never a gate)."""
+    """OCR street-name agreement with a candidate pose (signed: see evidence)."""
 
     score: float
     n_labels: int
     n_hits: int
     hits: list[dict] = field(default_factory=list)
+
+    @property
+    def evidence(self) -> float:
+        """The signed term the ranking uses (see name_evidence)."""
+        return name_evidence(self.score, self.n_labels, self.n_hits)
+
+
+def name_evidence_of(name: dict | None) -> float | None:
+    """Signed name evidence from a serialized name block, or None when absent.
+
+    Records written before issue #375 carry no ``evidence`` key, so the value
+    is recomputed from the fields that were always logged.
+    """
+    if not name:
+        return None
+    if "evidence" in name:
+        return float(name["evidence"])
+    score, n_labels, n_hits = (
+        name.get("score"),
+        name.get("n_labels"),
+        name.get("n_hits"),
+    )
+    if score is None or n_labels is None or n_hits is None:
+        return None if score is None else float(score)
+    return name_evidence(float(score), int(n_labels), int(n_hits))
+
+
+def name_evidence(score: float, n_labels: int, n_hits: int) -> float:
+    """Name agreement as SIGNED evidence: hits earn, unmatched labels cost.
+
+    ``score`` is name_alignment's reward-only value, Sum(exp(-d/tau)) over hits
+    divided by (n_labels + 2); this undoes that denominator, charges
+    W_NAME_MISS per eligible label that found no match, and restores it. A
+    pose that matches every label is unchanged, a pose that matches none goes
+    negative, and pages with fewer than NAME_MISS_MIN_LABELS eligible labels
+    keep the old reward-only behaviour.
+    """
+    if n_labels < NAME_MISS_MIN_LABELS:
+        return score
+    denominator = n_labels + 2
+    return (score * denominator - W_NAME_MISS * (n_labels - n_hits)) / denominator
 
 
 @dataclass
@@ -153,7 +206,7 @@ class SnapCandidate:
         """The ranking score: matcher verification plus soft evidence bonuses."""
         return selection_score(
             self.verification,
-            self.name.score if self.name is not None else None,
+            self.name.evidence if self.name is not None else None,
             self.region_containment,
             self.prior_theta_residual_sigma,
         )
@@ -161,21 +214,25 @@ class SnapCandidate:
 
 def selection_score(
     verification: float,
-    name_score: float | None,
+    name_evidence_value: float | None,
     region_containment: float | None,
     prior_theta_residual_sigma: float | None,
 ) -> float:
-    """The ranking score: matcher verification plus soft evidence bonuses.
+    """The ranking score: matcher verification plus soft evidence terms.
 
     One formula for the ladder's own candidates (SnapCandidate.select_score)
     and for synthetic poses scored after the fact (rank_pose), so a truth pose
     lands on the same footing as the search's candidates.
+
+    ``name_evidence_value`` is the SIGNED name term (name_evidence), not
+    name_alignment's reward-only score: a pose whose labels match nothing pays
+    for it rather than merely failing to earn.
     """
     if not math.isfinite(verification):
         return -math.inf
     score = verification
-    if name_score is not None:
-        score += W_NAME * name_score
+    if name_evidence_value is not None:
+        score += W_NAME * name_evidence_value
     if region_containment is not None:
         score += W_CONTAIN * region_containment
     if prior_theta_residual_sigma is not None:
@@ -790,6 +847,7 @@ def evaluate_pose(
         name = name_alignment(ctx.label_features, ctx.block_index, world_affine)
         evaluation["name"] = {
             "score": round(name.score, 4),
+            "evidence": round(name.evidence, 4),
             "n_labels": name.n_labels,
             "n_hits": name.n_hits,
         }
@@ -830,7 +888,7 @@ def rank_pose(
     evaluation["select_score"] = round(
         selection_score(
             evaluation["verification"],
-            name["score"] if name is not None else None,
+            name_evidence_of(name),
             containment,
             residual,
         ),
--- a/mapsnap/osm_snap.py
+++ b/mapsnap/osm_snap.py
@@ -74,13 +74,20 @@ W_NAME = 1.0
 W_CONTAIN = 0.3
 W_PRIOR = 0.1
 
-# Cost per eligible label that matches nothing at a pose (issue #375). The
-# reward half of name_alignment is floored at zero, so a pose whose labels all
-# land on the wrong ground scores the same as a page with no labels at all --
-# and a grid alias with strong P(road) shape evidence outranks the truth
-# (richmond p311: 0 of 9 labels hit, published 14,713 ft off, beating the
-# 18.8 ft pose by 0.046). Corpus-wide, zero hits with >=3 eligible labels
-# occurs in 0.8% of accurate poses and 43.3% of >=500 ft poses.
+# Cost charged to a pose whose labels match NOTHING (issue #375). The reward
+# half of name_alignment is floored at zero, so such a pose scores the same as
+# a page with no labels at all -- and a grid alias with strong P(road) shape
+# evidence outranks the truth (richmond p311: 0 of 9 labels hit, published
+# 14,713 ft off, beating the 18.8 ft pose by 0.046).
+#
+# The penalty is a TAIL, not a slope. What the corpus shows is that *zero*
+# hits with >=3 eligible labels is damning: it happens in 0.8% of accurate
+# poses and 43.3% of >=500 ft poses. It shows nothing about 1-of-5 being worse
+# than 3-of-5 in proportion, and a first cut that charged per miss regressed
+# exactly there -- nashville p24 (a correct pose hitting 4 of 7) lost 0.17 of
+# select score and stopped clearing PRODUCTION_GATE_SCORE, and brooklyn p9's
+# correct pose (1 of 5) was demoted under an alias that hits 4 of 5 because
+# sliding along a long avenue keeps a label near its own street.
 W_NAME_MISS = 0.5
 # Below this many eligible labels a page says nothing either way, so the
 # penalty stays inert (one unmatched label is noise, not contradiction).
@@ -153,19 +160,22 @@ def name_evidence_of(name: dict | None) -> float | None:
 
 
 def name_evidence(score: float, n_labels: int, n_hits: int) -> float:
-    """Name agreement as SIGNED evidence: hits earn, unmatched labels cost.
+    """Name agreement as SIGNED evidence: matching nothing costs.
 
-    ``score`` is name_alignment's reward-only value, Sum(exp(-d/tau)) over hits
-    divided by (n_labels + 2); this undoes that denominator, charges
-    W_NAME_MISS per eligible label that found no match, and restores it. A
-    pose that matches every label is unchanged, a pose that matches none goes
-    negative, and pages with fewer than NAME_MISS_MIN_LABELS eligible labels
-    keep the old reward-only behaviour.
+    ``score`` is name_alignment's reward-only value. A pose that matches at
+    least one of its eligible labels keeps that value unchanged; a pose that
+    matches NONE of them, on a page carrying at least NAME_MISS_MIN_LABELS
+    eligible labels, is charged W_NAME_MISS scaled by the same
+    n_labels/(n_labels + 2) shape the reward uses -- so contradiction grows
+    with how much the page had to say, and a page with nothing to say is
+    unaffected.
+
+    Deliberately not a per-miss slope: see W_NAME_MISS for the regressions
+    that shape produced.
     """
-    if n_labels < NAME_MISS_MIN_LABELS:
+    if n_labels < NAME_MISS_MIN_LABELS or n_hits > 0:
         return score
-    denominator = n_labels + 2
-    return (score * denominator - W_NAME_MISS * (n_labels - n_hits)) / denominator
+    return score - W_NAME_MISS * n_labels / (n_labels + 2)
 
 
 @dataclass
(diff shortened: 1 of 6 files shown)


danvk:
## Richmond A/B — +3.3, two rescues, zero losses

Control from a base-commit worktree at 102f287; `fit` only (OCR is untouched by this change), and `fit` clears the snap caches itself so both arms rank cold. The control reproduces this morning's independent Richmond control **exactly** at 83.2%.

| arm | score | ≤25 ft | 50–200 | **≥200 ft** | placed |
|---|---|---|---|---|---|
| control | 83.2% | 85.7% | 3.2% | **2.5%** | 90/92 |
| signed name, absolute penalty | 85.5% | 86.9% | 2.2% | **1.4%** | 89/92 |
| signed name, **page-relative** | **86.5%** | **87.8%** | 2.2% | **1.4%** | 90/92 |

Final arm: **exactly two pages move, both up**, 93 untouched.

| page | control | treatment | |
|---|---|---|---|
| p311 | **14,713 ft** | **18.8 ft** | disaster → good |
| p323 | 81.0 ft | **22.7 ft** | poor → good |

### The mechanism, confirmed in production

p311's candidate list after the change — the alias that was published is no longer even in the top three:

```
C0: rmse=    18.8  ver=0.229  name 0.410  evidence= 0.228  hits=5/9  sel=0.7925
C1: rmse=   262.6  ver=0.370  name 0.262  evidence=-0.011  hits=3/9  sel=0.6362
C2: rmse=   264.3  ver=0.370  name 0.262  evidence=-0.010  hits=3/9  sel=0.6357
```

p323, same story: the 2,372.6 ft pose scores **0 of 10 hits → evidence −0.4167**, dropping it behind the 22.7 ft pose (`sel` 0.2121 vs 0.3485) and flipping snap's verdict from `keep` to `refine`.

### Second commit: why the penalty had to become page-relative

The absolute form scored +2.3 but cost one page. p322's *correct* 13.5 ft RANSAC pose has **negative** P(road) verification (−0.5638) and paid 0.25 of miss penalty, moving its unary from −0.328 to −0.078 — close enough to the bar that a pairwise conflict tipped the joint solution to UNPLACED.

That exposed a real flaw rather than a tuning problem. `UNPLACED` is pinned at energy 0, so an absolute penalty does not merely re-rank a page's poses; it drags every hypothesis toward abstention. "This pose is wrong" and "this page is unplaceable" are different claims, and only the first is what name misses are evidence for.

`normalize_name_penalty` subtracts the page's *smallest* penalty from all of its hypotheses: ordering is preserved exactly (an alias keeps the relative part of its penalty), while the best hypothesis keeps the distance to UNPLACED that the reward-only calibration gave it. p322 comes back, both rescues stay, and the gate-recalibration worry largely dissolves for reconcile — the abstention bar sees the same numbers it always did.

### Still open

- **Snap's own gates** are still absolute (`PRODUCTION_GATE_SCORE = 1.35`); this A/B did not need them re-derived, but a page whose candidates are *all* contradicted now scores lower against them. Worth a look in a corpus run.
- **Single volume.** Richmond owns two of the 16 offline rescues; the other 14 sit in DC, brooklyn, columbus, fargo, KC, NO-1896, philadelphia, columbia, grand rapids and chicago. A corpus run is the real test.
- Weight sweep unchanged: 0.5 remains the largest miss cost with zero regressions offline.

1,088 tests pass; ruff and pyright clean.

danvk:
## Corpus A/B caught a design error: the penalty was the wrong shape

The single-volume richmond result was real but not representative. Running the corpus surfaced two things — one harness bug of mine, and one genuine error in the change.

### Harness bug first (results retracted)

`mapsnap fit` spawns every stage as a bare `["mapsnap", ...]`, resolved through **PATH**. Selecting the arm by invoking `$BASE/.venv/bin/mapsnap` therefore controls nothing: the stages follow the ambient PATH, which starts with the branch venv. The first five corpus volumes ran **treatment code in both arms** and reported a tidy row of "0 moved" nulls. It surfaced only because richmond's *control* scored 86.5% — the known treatment number. Each arm now exports its own PATH, unsets `VIRTUAL_ENV`, and logs `command -v mapsnap` into its own log. (The dedicated richmond A/B was never affected: `uv run` inside each project sets PATH correctly, which is why 83.2 vs 86.5 stands.)

### The real error: a tail statistic implemented as a slope

With arms actually isolated:

| volume | control | per-miss slope | delta |
|---|---|---|---|
| champaign | 97.6 | 97.6 | 0.0 |
| brooklyn | 83.6 | 82.9 | **−0.7** |
| nashville | 71.2 | 68.2 | **−3.0** |

Every damaged page has a signature the rescues lack:

| page | outcome | hits on the correct pose |
|---|---|---|
| richmond p311 | rescued | alias hit **0 of 9** |
| richmond p323 | rescued | alias hit **0 of 10** |
| brooklyn p9 | regressed | truth hit 1 of 5 — the **399 ft alias hit 4 of 5** |
| nashville p24 | regressed | truth hit 4 of 7 |
| nashville p4 | regressed | truth hit 2 of 3 |

The corpus statistic behind this PR (0.8% of accurate poses vs 43.3% of ≥500 ft poses) is entirely about the **zero-hit tail**. Charging `W_NAME_MISS` *per miss* additionally asserted that 1-of-5 is proportionally worse than 3-of-5 — which the data never showed, and which is false on brooklyn p9, where sliding along a long avenue keeps four of five labels near their own streets.

Nashville p24 also showed the gate flaw concretely: snap ranked the correct 15.2 ft pose **first in both arms**, but the slope cost it 0.17 of select score and it stopped clearing the absolute `PRODUCTION_GATE_SCORE = 1.35`. Normalizing inside reconcile did not help, because snap's own gates are absolute.

### The fix, and re-measurement

`name_evidence` now charges only a pose that matches **nothing**, scaled by `n_labels/(n_labels + 2)`. A pose with any agreement keeps its exact reward-only score, so absolute gates stay calibrated and partial matches are never taxed.

| volume | control | slope | **tail** |
|---|---|---|---|
| richmond | 83.2 | 86.5 | **86.5 (+3.3)** |
| nashville | 71.2 | 68.2 | **71.2 (0.0)** |
| brooklyn | 83.6 | 82.9 | **83.1 (−0.5)** |
| champaign | 97.6 | 97.6 | **97.6 (0.0)** |

Brooklyn's p9 regression is gone — its candidates now score identically to control. The residual −0.5 is p13 going unplaced → 634 ft, a coupling knock-on from p4 improving 7,668 → 261 ft (an improvement the metric cannot reward, since both ends are disasters).

Tests pin both regressions by name. The full 20-volume corpus A/B is running on the tail form.

danvk:
## Corpus A/B, 20 volumes: **+0.09 mean. Not worth merging as it stands.**

Control from a base-commit worktree at 102f287, treatment on this branch, both arms PATH-isolated with `command -v mapsnap` logged, volumes smallest-first, snap caches cold by construction. Sixteen of twenty controls reproduce their 2026-08-28 baseline exactly; the four that differ are #371's known gains.

| volume | control | treatment | delta | pages moved |
|---|---|---|---|---|
| richmond_va_1925_vol_3 | 83.2 | **86.5** | **+3.3** | p311 14,713→18.8; p323 81.0→22.7 |
| detroit_mich_1929_vol_11 | 77.2 | 76.3 | **−0.9** | p93 unplaced→293.7 |
| brooklyn_ny_1939_vol_1 | 83.6 | 83.1 | **−0.5** | p4 7,668→261; p13 unplaced→634 |
| chicago_il_1950_vol_1 | 80.0 | 80.0 | 0.0 | p1N 4,476→405 |
| miami_fl_1950_vol_1 | 83.3 | 83.3 | 0.0 | p74 32.8→unplaced |
| *15 others* | | | **0.0** | none |

**Mean delta +0.09** (corpus mean 80.46 → 80.55). One win, two losses, seventeen untouched. **Seven pages moved out of ~2,163.**

### What the corpus taught that one volume could not

**The wins are real but rare.** Richmond's two rescues are the whole positive signal, and both are the pure case the term was designed for: a grid alias that matches **zero** of nine or ten street labels, published at 14,713 ft and 81 ft. Chicago p1N (4,476→405) and brooklyn p4 (7,668→261) are the same mechanism working correctly — large corrections that the band metric cannot credit because both ends sit above 200 ft.

**The losses are a candidate-set effect, not an energy effect.** Detroit p93 and brooklyn p13 both go *unplaced → disaster*. Snap's top two candidates score identically in both arms; what changes is the third. Penalizing a zero-hit candidate evicts it from the top-K, a different (still wrong) pose takes the slot, and reconcile's hypothesis menu changes enough to clear the entry bar. My page-relative normalization in `unary_energy` cannot see this, because the damage happens before reconcile scores anything.

**Miami p74 is the predicted false positive**, and it landed exactly where the issue said it would: a *correct* 32.8 ft pose that matches 0 of 3 labels, charged −0.3 and dropped to unplaced. Miami owned 4 of the 14 accurate-but-zero-hit poses in the corpus census. It is score-neutral only by luck (32.8 ft was mid-band).

### Options

1. **Raise `NAME_MISS_MIN_LABELS` from 3 to 5.** Miami p74 (3 labels) stops being charged; richmond's rescues (9 and 10 labels) are untouched. Cheap to test and strictly narrows the blast radius.
2. **Keep the penalty out of the top-K cut** — rank with it, select candidates for reconcile without it — which addresses both losses directly, since neither is about scoring.
3. **Ship it as-is.** +0.09 with 7 pages moved is inside the noise; the honest reading is that the term fixes one spectacular failure and costs two smaller ones.
4. **Don't merge.** The evidence does not support a corpus-wide claim.

My recommendation is 1 + 2 measured together on the five volumes that actually move (richmond, detroit, brooklyn, chicago, miami) before any corpus rerun — a ~25 minute experiment rather than three hours. If that does not turn the two losses into nulls while keeping richmond, this should stay unmerged: a term that changes 7 pages in 2,163 is not paying for its risk.

Also worth recording: the next A/B of this should be **flag-gated in one build** rather than two venvs selected by PATH. `mapsnap fit` spawns each stage as a bare `["mapsnap", ...]`, so PATH decides which code runs, and an editable install means the working tree is live code — an in-flight run forbids touching the repo.

danvk:
## Label floor of 5: both losses gone, both rescues kept

The corpus A/B handed me an unusually clean diagnosis. **Every page it damaged carried exactly three eligible labels** — detroit p93 and brooklyn p13 (unplaced → disaster), miami p74 (a correct 32.8 ft pose matching 0 of 3, which lost its placement). **Both rescues carry nine and ten** (richmond p311, p323). So raising `NAME_MISS_MIN_LABELS` from 3 to 5 was predictable from the harvested records before running anything.

It is also what the evidence says on its own merits. Recomputed on **13,326 fresh candidates from this run's own control arms** — not the volume roots, whose records predate #371 and were the source of my original overstated projection:

| floor | accurate poses w/ zero hits | wrong poses w/ zero hits | ratio |
|---|---|---|---|
| ≥3 | 0.8% | 43.6% | 51:1 |
| **≥5** | **0.4%** | **39.4%** | **105:1** |
| ≥6 | 0.2% | 36.3% | 199:1 |

Thin-evidence pages are precisely where an accurate pose legitimately matches nothing, so the floor halves false positives while the true-positive rate barely moves.

### Re-measured on the five volumes that move at all

Treatment arms only, against the same controls:

| volume | control | floor 3 | **floor 5** | pages moved at floor 5 |
|---|---|---|---|---|
| richmond | 83.2 | +3.3 | **+3.3** | p311 14,713→18.8; p323 81→22.7 |
| detroit | 77.2 | −0.9 | **0.0** | none |
| brooklyn | 83.6 | −0.5 | **0.0** | p4 7,668→261 |
| chicago | 80.0 | 0.0 | 0.0 | none |
| miami | 83.3 | 0.0 | 0.0 | none |

**Corpus mean +0.095 → +0.165, with no volume regressing.** Brooklyn p4's 7,668 → 261 ft correction survives; the band metric cannot credit it because both ends are disasters. The top-K change I floated is not needed — both losses were thin-evidence pages, not a candidate-set problem that survives the floor.

A full 20-volume confirmation run at floor 5 is in flight. Floor 5 penalizes a strict subset of what floor 3 penalized, so the 15 untouched volumes are very unlikely to move — but "very unlikely" is not "measured", which is the whole lesson of this PR.

### Knobs are now runtime-selected

`W_NAME_MISS` and `NAME_MISS_MIN_LABELS` read `MAPSNAP_NAME_MISS` and `MAPSNAP_NAME_MISS_MIN_LABELS`, defaulting to 0.5 and 5. Environment rather than a CLI flag because `mapsnap fit` spawns every stage as a subprocess: the environment crosses that boundary, where a flag would need threading through each stage. One build, arm chosen at run time — no more two-venv PATH games, which is how this PR's corpus run got contaminated in the first place.

danvk:
## Confirmation run (partial, stopped by request) and where this stands

Four volumes of the floor-5 corpus run, then stopped once it behaved as predicted:

| volume | control | treatment | delta |
|---|---|---|---|
| richmond_va_1925_vol_3 | 83.2 | **86.5** | **+3.3** |
| brooklyn_ny_1939_vol_1 | 83.6 | 83.6 | 0.0 (p4 7,668→261, band-neutral) |
| champaign_ill_1915 | 97.6 | 97.6 | 0.0 |
| nashville_tn_1957_vol1a | 71.2 | 71.2 | 0.0 |

Combined with the treatment-only re-runs of detroit (0.0), chicago (0.0) and miami (0.0), every volume that moved under floor 3 has now been re-measured at floor 5, and the projected corpus position is **+0.165 mean with no volume regressing** (the other 13 were bit-identical at floor 3, and floor 5 penalizes a strict subset).

## Merge recommendation, stated honestly

**The measured value is one volume, and that volume has an upstream root cause.** Richmond p311 is a disaster because a General-Index "1" is misread as page "311", seeding snap's search 4.5 km from the sheet (the case study on the #270 docket). Fixing the keymap read fixes p311 without touching pose scoring — and if that lands, this PR's only *scored* win goes with it.

**What survives that argument** is that the term is a guard against a class, not a patch for two pages. On pages carrying ≥5 eligible labels, **39% of ≥500 ft poses match zero street names, against 0.4% of accurate poses** — a ~105:1 likelihood ratio the ranking currently discards entirely. Today the gates and reconcile catch most of those before publication, which is why only three pages move corpus-wide. That is a statement about the pipeline's *current* defenses, not about the risk: #270 phase 3 plans to delete the stage gates in favour of joint arbitration, and that endgame needs a name channel that can say **no**, not only yes.

So: merge it for the asymmetry it removes, not for the +0.165. If you would rather not carry the concept until reconcile actually needs it, closing this unmerged and leaving #375 open with the corpus evidence attached is equally defensible — `scripts/name_evidence_sim.py` and the fresh-record census make it cheap to revive, and the branch is a complete record.

The one framing to avoid is merging this **as the fix for p311**. It is a downstream mitigation of a keymap misread, and the honest fix for that page is upstream.

## What is on the branch

- `name_evidence`: a pose matching **nothing** on a page with ≥`NAME_MISS_MIN_LABELS` (5) eligible labels is charged `W_NAME_MISS` (0.5) scaled by `n_labels/(n_labels + 2)`; any pose with agreement keeps its exact reward-only score, so absolute gates stay calibrated.
- `normalize_name_penalty` in reconcile: the penalty is relative within a page, so it re-ranks poses without dragging the whole page toward UNPLACED.
- Records carry `name.evidence`; `name_evidence_of` recomputes it for pre-#375 records.
- Both knobs read the environment, so a sweep is one build with the arm chosen at run time.
- `scripts/name_evidence_sim.py` reproduces the offline analysis from cached candidates.
- 1,090 tests; the two corpus regressions (brooklyn p9, nashville p24) and the thin-evidence class (miami p74) are pinned as named test cases.

danvk:
This is lower-impact than I'd hoped, but it's conceptually reasonable (if you can match lots of street names, you should!) and is a good defense against some real head scratchers that can come from bad location priors.

Issue #276, opened 2026-08-09
Keymap sheets: detect and mask the "Graphic Map of Volumes" inset


danvk:
Follow-up to the #83 splitter work, feeding #211 (keymap P(road)) and #261 (keymap trust). Surveyed all 20 keymap sheets across the 17 volumes: **6 carry a "Graphic Map of Volumes" inset — a second map of the whole city at coarse scale, with its own numerals and roads.** The splitter separates it on 2 of the 6.

(The other 14 sheets — brooklyn ×2, chicago, columbia, fargo, grand_rapids ×2, hudson, nashville, NO-1896, philadelphia, DC, LA pb — have none; multi-volume references there are marginal text like "SEE VOLUME SIX". Champaign has a small *state locator* inset, a different beast. Chicago vol 1, a deeply multi-volume city, has no graphic inset at all.)

## Survey

| sheet | inset position | bounded how | splitter outcome |
|---|---|---|---|
| kansas_city p0 | bottom-left | boxed; borders 6.0 px thick, both run to sheet edges | **✓ split out** (`p0__2`) |
| new_orleans_1951 p0 | top-left | boxed; same corner geometry | **✓ split out** (`p0__1`) |
| detroit p0 | top-left | **unboxed** — banner + free-floating map | ✗ sheet *is* split, but the cut panel is the KEY-TO-SYMBOLS box; the volume map stays inside the keymap panel `p0__1` |
| columbus p0 | top-left | thick divider with 45° chamfered corners (pasted patch) | ✗ unsplit |
| miami p0 | bottom-right | **unboxed** — floats beside the KEY box | ✗ unsplit |
| los_angeles pa | bottom-right | **unboxed**; plus a second inset (Vol-13 continuation) bounded by a thick irregular border | ✗ unsplit |

Detroit is a near-miss that flatters the stats: the splitter found the one genuinely boxed region on the sheet (the key legend), so the keymap panel still contains the volume map.

## Why the current splitter can't do better (traced per stage, measured)

1. **LA pa never gets started: the dense-grid bail.** 14 long candidates > `MAX_DIVIDER_CANDIDATES = 12`, so `connected_dividers` returns `[]` by design — a keymap *is* the dense grid that cap exists to reject.
[...]

PR #387, opened 2026-09-02
Consumers of key-map reads skip the ones inside a volume-index inset

danvk:
Step 4 of the #276 plan: the masks from #386 are now consumed.

## Mechanism

`mapsnap keymap-inset` flags every read inside a confirmed inset mask with **`inset: true`** in the sheet's `<stem>.keymap.json`. The flag is recomputed from scratch each run (a mask that goes away clears its flags), and the reads stay in the file so the debugger shows what was masked. Every consumer of page-number reads then skips flagged records through one predicate, `records.is_inset`:

| consumer | effect |
|---|---|
| `fit_keymap.load_detections` — the loader behind `KeymapLocator` | no location, no search center (`centers_for`), no radius contribution, no key-map GeoJSON entry |
| `page_regions.load_seeds` | an inset's numerals seed no page region |
| assignment-repair planner (`planner_inputs`) | neutralized **in place** — empty label, no center — because `apply_repairs` indexes into the same list; it cites nothing, is cited by nothing, is never relabelled |
| `road_prob.mapped_extent_mask` | the inset is cut out of the supervised extent: its roads are real but drawn at another scale |

The key-map pipeline order already puts the inset step before repairs and georef, so the flags are in place when they are read.

## Richmond p311, end to end

The case that started #276. Georeferenced in a scratch volume against the key map with the flags stripped, then with them in place:

| | key-map search centers | distance from truth | radius | GCPs |
|---|---|---|---|---|
| before | **3** | 0.0, 1.1, **4.9 km** | 569 m | 2 |
| after | **2** | 0.0, 1.1 km | 561 m | 2 |

The inset's "1" — snapped to "311" by the vocabulary repair — no longer seeds a snap search 4.9 km away. The real centers and the GCPs are untouched.

On the seven inset-bearing sheets the flags land on exactly the reads the detector reported: richmond `2 3 4 311`, miami `5 5 4 2 5`, detroit `15 13 8 10`, columbus `5 1 4A 4 4 4 5 6A`, los_angeles `4 1 3 3`; kansas_city and new_orleans_1951 have none to flag.

## Not here

Masking the inset at P(road) *inference* (zeroing the prediction inside the rings) — the training extent is handled; the prediction consumer in the key-map snap path is a separate, small follow-up.

Tests: the predicate, the annotation round-trip (flags set, then cleared when nothing confirms), and each consumer's skip. 1,132 tests, ruff and pyright pass.

Part of #276 (step 4 of the plan there).

### DIFF
--- a/docs/fit-pipeline.md
+++ b/docs/fit-pipeline.md
@@ -13,7 +13,9 @@ key map), `detect-numbers` (candidates, reads, valid page keys not read),
 `assignment-repair`, `cartouche` (the GRAPHIC MAP / KEY / INDEX words read and
 where), and `inset` (each isolated small-number cluster, its no-snap re-read,
 the cartouche words near it, and the verdict; the confirmed masks are written
-to `raw/<stem>.inset.panels.json`). A rerun replaces a stage's section in
+to `raw/<stem>.inset.panels.json`, and the reads inside them are flagged
+`inset: true` in `<stem>.keymap.json` — kept for the debugger, skipped by every
+consumer of page-number reads). A rerun replaces a stage's section in
 place, each header carries a UTC timestamp, and `archive_run` copies the logs
 into `<run>/raw/`.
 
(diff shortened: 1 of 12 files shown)


danvk:
This looks pretty solid. Here's Richmond:

image

PR #391, opened 2026-09-02
Snap panel shows how select and verif are built, on hover

danvk:
The snap explorer showed the `select` and `verif` scores but not what goes into them, so the street-name term from #375 (and containment, and the rotation-prior residual) were invisible even though every candidate record carries them.

image

Hovering a `select` or `verif` cell now pops a breakdown over the table: each term, its weight and input, and the sum against the recorded value.

```
select = verif + 1.0 × name.evidence + 0.3 × containment + 0.1 × max(−1, 1 − prior residual σ)
verif  = inlier_frac + ncc_fine − chamfer_mean_m / 30 m
```

- A term the record lacks is listed as missing and contributes nothing, as the pipeline skips a `None` term.
- A name block without the signed evidence (records before #375) falls back to the reward-only score and says so. Richmond p365's `ab-incumbent-rung` record is one such.
- When the terms do not add up to the recorded score, an amber note says by how much. The weights live in the pipeline (`osm_snap.W_NAME` / `W_CONTAIN` / `W_PRIOR`, `edge_join.CHAMFER_CLAMP_M`) and are mirrored in `snap.ts`; this note is how a change there shows itself in the debugger.

**UI care.** The popover is fixed-positioned from the cell's viewport rectangle and opened by hover state, so it escapes the snap panel's `overflow: auto` clipping and never reflows the rows: selecting a row does not make the table jump. It flips above the cell when there is no room below. Verified live on Richmond p365 in the running debugger: the panel's scroll height is unchanged with the popover open, and the four terms add back to the recorded 2.2823.

The breakdown helpers (`selectBreakdown`, `verificationBreakdown`) are pure and unit-tested; the panel test checks every row carries both tooltips and that the mismatch note fires on a record with missing terms.

### DIFF
--- a/app/src/components/SnapPanel.test.tsx
+++ b/app/src/components/SnapPanel.test.tsx
@@ -127,6 +127,23 @@ describe('SnapPanel phase-2 records', () => {
     );
   });
 
+  it('carries each score breakdown in a hover popover, not in the row', () => {
+    const html = render(record);
+    // Every row's select and verif cells carry a breakdown tooltip.
+    expect(
+      (html.match(/class="snap-breakdown" role="tooltip"/g) ?? []).length,
+    ).toBe(8);
+    // The truth row's name term, from its recorded alignment block.
+    expect(html).toContain('5/9 labels hit');
+    expect(html).toContain('record predates the signed term');
+    expect(html).toContain('no key-map region for this page');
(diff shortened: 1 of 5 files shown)


danvk:
I don't love that this duplicates formula logic in the frontend, but we can fix that when it becomes an issue.

Issue #352, opened 2026-08-26
Content-region overlap as a wrong-fit signal and gap-filling constraint


danvk:
Investigating a second use for the #226 region model: pages' content regions should **tile** the ground rather than overlap, so overlap between two placed pages is evidence that one of them is misplaced — and the uncovered gap between placed neighbours is where an unplaced page belongs.

## The premise holds, strongly

Placing each page's **truth** content mask at its **truth** pose and measuring pairwise intersection-over-min-area:

| volume | pages | pairs | >1% | >10% | max | own region covered by all others (median) |
|---|---|---|---|---|---|---|
| detroit | 99 | 4,851 | 10 (0.2%) | **0** | 0.03 | 0.1% |
| hudson | 89 | 3,916 | 7 (0.2%) | **0** | 0.05 | 0.3% |

Content regions are essentially disjoint: no pair anywhere overlaps by more than 5%, and a typical page has 0.1–0.3% of its own region covered by every other page combined. For contrast, the *full page rectangles* at the same poses overlap by >10% in 2.7% (detroit) and 5.9% (hudson) of pairs — the margins and duplicated content are exactly what the mask removes.

~~**NO-1896 cannot participate**: its OIM annotations mark the whole sheet (truth region/page area ratio **1.00**, vs detroit 0.72 and hudson 0.50), so it has no content-region signal in truth. Worth knowing separately — the region model is trained on a mix of masked and whole-page volumes, which likely contributes to its over-prediction tail.~~

**Correction (2026-09-08):** the whole-sheet selectors were a stale truth export, not an OIM gap. OIM's Main Content multimask for NO-1896 vol 2 has 91 features (last saved 2026-08-06); our `main.iiif.json` dated from 2026-08-01 and every export in its lineage predated the mask, so 76 of 91 selectors were the full-canvas rectangle OIM emits for an unmasked layer. The truth was re-exported on 2026-09-08 (92 items, 91 masked; only p125 is unmasked on OIM) and `fix-truth-splits` applied, so NO-1896 **can** participate. Regrading the 2026-09-03 run against the new truth moves the volume 78.7 → 76.7 (−0.9 from the newly georeferenced p122, unplaced in that run; −1.1 from OIM GCP edits on 10 pages).

[...]

PR #401, opened 2026-09-08
Per-page content-region maps and a Page / Region / P(road) toggle in the volume viewer

danvk:
## What

- **`mapsnap region data/<vol>`** (new): runs the content-region UNet from #355 over every 25% page image, parents and split panels alike, and writes each prediction as an 8-bit P(region) PNG at the page's own resolution to `data/<vol>/artifacts/region/<stem>.png` (0 = margin, 255 = content). Existing maps are kept unless `--force`. Detroit's 116 pages take 17 s on MPS. Until now nothing consumed the region model; its only output was a JET preview overlay.
- **Volume viewer toggle**: a Page / Region / P(road) segmented control under the opacity slider (URL `pages=region|roadprob`). The annotation route accepts `?image=` and points each page's image service at the PNG instead of the sheet, with the PNG's real dimensions, so the maps warp and clip exactly as the sheets do. Pages without the map fall back to their sheet and the status line says how many (`86 pages have no region map (mapsnap region <volume>), drawn as sheets`). P(road) reuses the page maps the snap experiments leave in `artifacts/edge_join/roadprob/`.
- `load_region_model()` factored out of the region model's predict command; a PNG IHDR dimension reader beside the JPEG one; README notes for the command and the viewer.

## Why

#352 measures content-region overlap as a wrong-fit signal but there was no way to look at the regions. On Detroit's 2026-09-03 run, placing the predicted regions at the published poses shows all three ≥200 ft disasters overlapping a neighbour by 32–49% of the smaller region (p61 ↔ p59, p85 ↔ p86, p93 ↔ p94) while good pages sit at a median 3%. The toggle makes that visible directly on the map.

Part of #352.

## Verification

- `uv run pytest mapsnap/region_predict_test.py` (stub model: page-resolution PNGs, skip-existing, `--force`, unreadable input) plus ruff / pyright repo-wide.
- `npm run type-check`, `npx vitest run` (282 tests incl. the alternate-image rewrite, PNG reader, URL param round trip), `npx prettier --check .`
- Viewed Detroit 2026-09-03 in all three modes and NO-1896 (no region maps) for the fallback status.

### DIFF
--- a/README.md
+++ b/README.md
@@ -504,6 +504,8 @@ npm run dev
 
 Then visit localhost:5173/mapsnap/.
 
+The same app also hosts the **volume viewer** (`localhost:5173/mapsnap/?view=iiif`), which draws a whole run's pages warped onto OpenStreetMap; it needs the local API server (`npm run server`) alongside `npm run dev`. Its Page / Region / P(road) toggle draws each page as its sheet, its content-region map, or its road-probability map. Region maps come from `mapsnap region data/<vol>` (written to `artifacts/region/`), P(road) maps from the snap experiments' `artifacts/edge_join/roadprob/`.
+
 To deploy the debugger:
 
 ```
--- a/app/server/api.ts
+++ b/app/server/api.ts
@@ -14,6 +14,7 @@
 import type { Endpoint, GetEndpoint } from 'crosswalk/dist/api-spec';
 import type {
   GeorefAnnotationPage,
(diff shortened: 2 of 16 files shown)


danvk:
The region model would probably be better as an overlay, rather than a replacement, but this is OK as a start to get some visibility into it.

Issue #354, opened 2026-08-28
Run mapsnap over all volumes in the Library of Congress


danvk:
There are 9,046 volumes and 347,659 images.

General approach:

- Download all images via torrent and convert them to 25%-scaled JPEGs (not JPEG2000). Keep full-resolution versions of candidate keymaps (p0* or p1). This should be done streaming to avoid multi-terabyte storage requirements. Results go on S3. This can run on spot instances on AWS Batch. I might also want to run `mapsnap split` as part of this.
- Produce per-county, per-city dumps of OSM roadmaps. This is going to be a project in its own right.
- Run `mapsnap craft` against all the 25%-scale images using a dedicated GPU machine. Claude suggests a g4dn. Save all the resulting boxes.json files to S3. This only has to be done once, unless I update the splits. 
- Also on a GPU, run `keymap`: detection, OCR, georeferencing, CNN, CRNN, regions.
- Also on a GPU, run OCR on remaining pages.
- On spot instances, run `fit` on all the volumes.

Claude's cost estimate was $100-250 of compute (CPU+GPU) plus ~$10/month of storage costs.

danvk:
Fleshing out each stage with concrete numbers. Grounding data: the catalog's own distribution, measured per-page artifact sizes from hudson (25% jpg mean **864 KB**, boxes.json 28 KB, streets.json 59 KB, georef.json 6 KB), and measured stage timings from this laptop's corpus runs, derated to cloud hardware.

## Corpus shape (from sanborn.csv — this changes the plan's economics)

| percentile | images/volume |
|---|---|
| p50 | **21** |
| p75 | 57 |
| p90 | 100 |
| p99 | 127 (max 174) |

**58% of volumes have ≤25 images** (they hold only 79k of the 348k images), while the 907 volumes ≥100 images hold 103k. Consequences: the median fit job is minutes, not hours — ideal for spot; per-volume fixed costs (OSM extract, keymap chain) are amortized over few pages, so keymap/OSM work is a *larger* share of cost than the truth corpus suggests; and small towns mean sparser OSM and likely coarser sheet scales (#351's census question rides along for free — the family-scale rung of every fitted volume falls out of this run).

## Stage 0 — pilot (do this first)

[...]

PR #404, opened 2026-09-09
mapsnap loc-mirror: build the Sanborn mirror from LoC full-resolution JP2s

danvk:
Initial step towards #354

## What

`mapsnap loc-mirror MAPPING.tsv --jp2-dir … --out-dir … --staging-dir … --bucket s3://mapsnap-sanborn --upload` builds the whole-collection Sanborn dataset from the Library of Congress full-resolution JP2s served by the torrent's HTTP mirror. Every sheet's identity is its own filename (`06246_1914-0051`), so no sequence-number mapping is involved; the plan is a mapping TSV built from the torrent listing and the loc.gov catalog.

Three stages run concurrently, each bounded by a different resource:

1. **download** (threads, `--streams`, the mirror saturates near 8): each kept sheet's JP2 into `--jp2-dir`, mirroring the torrent tree and verified against the listed byte count, so the copy is kept as a valid torrent payload;
2. **decode** (processes, `--decode-workers`): `opj_decompress -r 2`, the JPEG 2000 quarter level, which is exactly the pipeline's 25% working scale, to `<staging>/by-state/<state>/<year>/<item>/p<key>.jpg` at quality 95 (what `mapsnap scale` writes); page-0 sheets and letter pages, the key-map candidates, also at full resolution to `raw/p<key>.jpg`;
3. **upload** (two `aws s3 sync` at a time): the item directory to `<bucket>/by-state/<state>/<year>/<item>/`, then its staging copy is deleted (`--keep-staging` retains it).

Up to `--prefetch-items` items download ahead of decoding so one-sheet volumes do not starve the connections. State lives only on local disk under `--out-dir`: per item `metadata.json` (catalog fields plus the sheet table: sequence, stem, key, source, bytes, dimensions, raw flag) with `.done` and `.uploaded` markers, plus `broken.log` (item, stem, source, reason) and `progress.jsonl`; resume never lists S3. Sheets without a numbered page key (covr, ind1, cbd, titl, note) are skipped. The 161 TIFF-only masters and 17 tile.loc.gov-only sheets are handled through the mapping's `source` column. One tqdm bar (`smoothing=0`) counts decoded sheets with downloaded GB, uploaded items, and broken sheets in the postfix.

## Verification

- `uv run pytest mapsnap/loc_mirror_test.py`: keep/candidate rules, mapping load, layout vs the S3 prefix, quarter and full decode of a generated JP2, the pipeline over a `file://` mirror with a corrupt sheet (outputs, metadata in both staging and state, broken log, resume with the mirror unreachable), and the upload stage with a stubbed `aws` (marker, staging removal, second run idle).
- Real runs: Alaska (27 items / 92 sheets / 0.7 GB in 104 s, resume 5 s) and Baja California with `--upload` (4 items / 32 sheets in 35 s, 36 objects in the bucket).
- ruff, pyright clean.

### DIFF
--- a/mapsnap/cli.py
+++ b/mapsnap/cli.py
@@ -83,6 +83,10 @@ SUBCOMMANDS: dict[str, tuple[str, str]] = {
     "download-osm": ("mapsnap.download_osm", "Download street data from OSM"),
     "osm-to-geojson": ("mapsnap.osm_to_centerlines", "Convert OSM data to GeoJSON."),
     "scale": ("mapsnap.scale_images", "Shrink images by a uniform amount."),
+    "loc-mirror": (
+        "mapsnap.loc_mirror",
+        "Build the Sanborn mirror (25% JPEGs + raw key-map candidates) from LoC JP2s.",
+    ),
     "download-oim": (
         "mapsnap.download_oim_iiif",
         "Fetch all images for a Sanborn volume from OldInsuranceMaps.net",
--- a/mapsnap/cli_test.py
+++ b/mapsnap/cli_test.py
@@ -24,14 +24,42 @@ def test_every_command_module_imports(name: str) -> None:
     the cycle is imported first, which is precisely what running the command
     does.
     """
-    for module_name in [key for key in sys.modules if key.startswith("mapsnap")]:
-        del sys.modules[module_name]
+    evicted = {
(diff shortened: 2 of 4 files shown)


danvk:
I'm about a quarter of the way through this, uploading to `s3://mapsnap-sanborn`

Issue #361, opened 2026-08-29
Invert split and craft: detect once on full pages, derive panel boxes at split time


danvk:
Today the pipeline runs `split` and then `craft`, so panel images are what CRAFT detects on — and any change to the splitter invalidates the corpus's most expensive vision pass. Proposal: **invert the order**. Run CRAFT once per full page, and have `split` *derive* each panel's `boxes.json` by remapping the parent's boxes (per-angle unrotate → crop offset → re-rotate, then panel assignment). CRAFT becomes truly run-once per corpus image; splitter iterations cost only CPU. This matters most for the all-LoC run (#354), where the CRAFT pass is 125–190 GPU-hours and the splitter is the component most likely to need iteration on unseen layouts.

## What inversion would cost — measured

63 split parents across schenectady / fargo / kansas_city / hudson: 21,092 parent boxes vs 20,585 panel boxes, compared per rotation angle in a common coordinate frame (mapping validated: mismatch rates are uniform across 0°/90°/270°; a bookkeeping error would explode the rotated angles).

**1. Seam-crossing boxes: negligible.** 28/21,092 parent boxes (**0.13%**) put ≥10% of area into two panels. Majority-panel assignment covers them; the residual cost is a clipped label or two per volume.

**2. Detection differences are real, modest, and favor the parent.**

| direction | count | rate |
|---|---|---|
| panel-only boxes (lost by inverting) | 1,400 | 6.8% of panel boxes |
| parent-only boxes inside panels (**gained** by inverting) | **1,912** | ~9% of parent boxes |

Panel-only rates split by panel type: full-canvas (masked) panels 2.8–4.9%, cropped panels 7.9–11.6% — the driver is crop scale/context. Detection is context-sensitive in both directions, and the full page is the slightly richer context: the current split-then-craft flow is *already* dropping ~1,900 boxes the parent images find.

**3. Consequence in reads: small.** Of 723 substantive vocab-matched reads on these panels (len ≥4, conf ≥0.5), **26 (3.6%)** trace to panel-only boxes — a mix of junk ("EW PIPE") and real names ("MILLARD ST", "17TH"). The 1,912 gained boxes are reads the current flow never attempts, an unmeasured upside of comparable or larger size.

## Design sketch

- `craft` runs on parent pages only (and keymap raws, unchanged). `split` gains a remap step: for each panel, transform the parent's per-angle boxes into the panel frame and write the panel's `boxes.json` (tagged `derived_from_parent: true` for provenance).
[...]

PR #412, opened 2026-09-14
Cache P(road) as a sidecar beside each page, predicted before split

danvk:
`mapsnap roadprob` is the road-UNet counterpart of `mapsnap craft`: a pass over parent sheets writing `<stem>.roadprob.jpg`, whose output depends only on the image and the model. `split` then cuts a panel's map out of its parent's, the way panel CRAFT boxes are already derived from the parent's (#361), so both heavy GPU stages run before the split and never again — the GPU tier of the corpus run becomes two stateless passes over immutable inputs, and everything after it is CPU work.

Part of #354.

## Is the derived map the same as a directly predicted one?

Yes, to within the UNet's own tiling noise. Measured on 30 panels from Schenectady, Kansas City and DC, comparing P(road) predicted on the parent and cropped against P(road) predicted on the panel image:

| comparison | median mean abs diff | median road-mask IoU |
|---|---|---|
| parent-then-crop vs direct, whole panel | 0.013 | 0.93 |
| same, interior (>64 px from any edge) | 0.008 | 0.95 |
| same, 64-px edge band only | 0.03–0.13 | 0.60 |
| noise floor: same panel, tile grid shifted 32 px | 0.014 | 0.92 |

They diverge only along the cut, where the parent-based map has the neighbouring panel's real ink in its receptive field and the panel-based one sees the white fill `split` painted there — so the derived map is the better of the two. `crop_to_panel` repeats `split`'s own bounding-box rounding (truncate low corner, round high corner) so the map lands in exactly the panel's pixel frame; a test asserts that against `write_panels`' real output, since a pixel of drift would offset everything `snap` matches.

The region UNet is deliberately left where it is: it letterboxes the whole page to 512 px, so a parent prediction is not a panel prediction, and it costs 0.3–0.6 s/page on CPU anyway.

## Storage

Quality-90 JPEG: 158 KB per 25% page against 684 KB for the equivalent PNG, at a mean error of 0.0008 (0.2 of one 8-bit level), far below anything the chamfer/NCC matching responds to. That is 65 GB rather than 282 GB over the corpus.

## Consumers

All readers go through `roadprob.load_roadprob`, which falls back to the pre-#354 `artifacts/edge_join/roadprob/<stem>.png`, so volumes fitted before this keep working with no re-run; `snap` still infers on demand for a page with no map. Updated: `snap`/`osm_snap_experiment`, `edge_join_experiment`, `keymap-snap`, `georef_from_labels`, the three pipelines (a `roadprob` step after `craft`), `docs/HOW-IT-WORKS.md`, the README, and the debugger's volume viewer — its P(road) toggle now reads the sidecar (with JPEG dimensions) and tries the legacy PNG second.

Key-map sheets are untouched: their P(road) comes from the colour model and already lives beside the sheet at `raw/<stem>.roadprob.png`.

## Also: `ocr --no-gpu` never worked on x86 Linux

Found while benchmarking. EasyOCR dynamically quantizes the recognizer to int8 on CPU. That silently no-ops on Apple Silicon (torch's quantized engine is "none") but succeeds on x86 Linux, where loading the fine-tuned float weights then fails on a state-dict mismatch — every `--no-gpu` ocr run on EC2 died there. Readers are now built unquantized, which also makes CPU and GPU workers recognize identically, and pool workers get `cpu_count // workers` torch threads rather than all cores each (8 workers on 8 cores had measured no faster than 1).

## Verified

- ruff format, ruff check, pyright, pytest (1188) repo-wide; `tsc --noEmit` and vitest (106) in `app/`.
- End to end on Kansas City p451: `mapsnap roadprob` wrote the parent map and derived both panels, whose frames match the panel JPEGs exactly (1949x1638 and 444x313); `--resume` is a no-op on the second run.
- `load_prob` resolves new sidecars and the legacy PNGs of an already-fitted volume (Hudson p49).


danvk:
I haven't tested this at all, but it seems innocuous (similar to inverting the order of craft and split) and it will greatly simplify #354 since it will mean I can run `craft` and `roadprob` on dedicated GPU machines before running the rest of the pipeline on plain old CPUs.

PR #414, opened 2026-09-14
loc-craft: make a run measurable, and let --limit mean work done

danvk:
Two small changes the first pilot needs, both found while writing its run book. Part of #354.

- **Progress lines carry a timestamp and the shard number.** `--workers N` tees several driver processes into one log, so without attribution the arms of a throughput comparison cannot be separated.
- **The summary reports elapsed seconds and pages per hour.** Items range from 1 to over 100 pages, so items per hour does not compare two configurations; and `h:mm` alone is too coarse for a ten-minute pilot.
- **`--limit` counts items processed, not items seen.** It included skips, so re-running a pilot on a shard that had already done that many items stopped having done nothing — exactly the loop a pilot iterates through.

The README now carries the A/B recipe: two instances on disjoint shards, one at `--workers 1` and one at `--workers 2`, compared on pages per hour.

Verified: ruff, pyright, 1203 tests. The new test drives `main()` through a dry run where two of three items are already complete and asserts `--limit 1` still processes the third.


danvk:
The goal here is to determine whether `--workers 2` is a significant win over `--workers 1` on the GPU spot instances for the initial craft + roadprob step. Next step is to run an A/B test to validate the workflow and answer that question.

PR #426, opened 2026-09-14
Add osm-counties: cut one OSM extract per Sanborn county from a national dump

danvk:
The corpus needs street geometry for **2,714 counties covering 432,410 sheets**. Fetching each from an OSM service would be thousands of requests; cutting them from one dump is a single pass per batch. Part of #408.

```sh
mapsnap osm-counties --counties counties.tsv --natural-earth ne_10m_admin_2_counties/ \
  --pbf us-named-highways-260826.renumbered.osm.pbf --out-dir /Volumes/fivetera/osm-by-county
```

Counties come from `loc-counties`' `counties.tsv`, boundaries from Natural Earth keyed on FIPS — all 2,714 resolve. `--natural-earth` takes a GeoJSON, a `.shp`, or the directory, converting with `ogr2ogr` when needed. Outputs are `<FIPS>.osm.pbf`, and a county already on disk is skipped so an interrupted run resumes.

## Two details worth review

**The buffer is latitude-aware.** Each boundary grows 3 km before cutting, because a town on a county line needs the streets across it and Natural Earth's outline is generalized anyway. The buffer is applied in a local metre frame scaled by the county's own latitude: a degree of longitude is 111 km at the equator and 85 km at 40°N, so buffering in raw degrees would stretch northern counties east-west. Two tests pin this, including one asserting a 60°N county grows wider *in degrees* than a 40°N one.

**The dump must be renumbered**, and the docstring says why so nobody rediscovers it: osmium sizes a node-id bitmap per extract from the highest id it sees, so raw OSM ids cost about 1.4 GB per polygon — a 100-polygon batch wanted 141 GB, was killed, and left truncated but structurally valid output files. Memory still grows with batch size even renumbered, so batches are capped (default 200) rather than run as one pass.

## Measured

On 20 counties against the real dump:

| | |
|---|---|
| time | 38 s |
| peak memory | 1.03 GB |
| Cook County, IL | 722,440 nodes, 122,149 ways, 11 MB |
| Los Angeles County | 20.8 MB |
| known Chicago street names | 691 hits |

ruff, pyright, 1244 tests.


danvk:
"The dump must be renumbered" is unfortunate since it means we lose the real OSM IDs. Maybe there's a way to get the original IDs and efficient outputs, but this is fine for now. I've copied these extracts to, for example, `s3://mapsnap-sanborn/osm-by-county/US18029.osm.pbf`. It's ~1.8GB for the 2,714 counties.

PR #430, opened 2026-09-15
Map the independent-city Sanborn items to OSM boundary relations

danvk:
`mapsnap loc-counties` keys every Sanborn item to a Natural Earth county, which is the unit `mapsnap osm-counties` cuts street extracts for. 428 items covering 13,665 sheets fall through, because their LoC county reads "independent cities": Virginia's cities belong to no county at all, and neither do Baltimore or St. Louis, so Natural Earth has nothing to key them on.

OSM does carry them, as `admin_level=6` relations tagged `border_type=city` — the county tier, because that is where they sit. Most also carry `nist:fips_code`, which is the identifier `loc-counties` already writes, so a matched city joins the existing pipeline instead of starting a second naming scheme.

## Result

| | items | sheets |
|---|---|---|
| matched by state + name | 406 | 13,531 |
| redirected to a successor | 22 | 134 |
| **total** | **428** | **13,665** |

36 distinct relations, covering 33 cities plus 3 counties that absorbed a former one. No overlap with the existing `items.tsv`, so the two files concatenate.

The five cities with no city relation are historical rather than mistagged, and each is redirected to whatever covers its streets now:

| city | redirected to | why |
|---|---|---|
| Warwick | Newport News | merged in 1958 |
| Fortress Monroe | Hampton | Fort Monroe, absorbed |
| Bedford | Bedford County | a town again since 2013 |
| Clifton Forge | Alleghany County | a town again since 2001 |
| Colonial Beach | Westmoreland County | a town, never independent |

Those rows are marked `successor` in the `match` column so they stay distinguishable from a name match.

## Verification

The boundary extract was checked rather than assumed: all 371 member ways and all 22,708 of their nodes are present, and all 36 relations assemble into polygons under `osmium export`. The largest are Bedford County, Alleghany County and Suffolk, which is right.

```sh
mapsnap loc-cities --skipped skipped.tsv --items items.tsv \
    --pbf us-counties-260826.osm.pbf --out-tsv city-items.tsv \
    --out-pbf independent-cities.osm.pbf
```

Part of #354.


danvk:
independent-cities.osm.pbf looks pretty reasonable in JOSM:

image

PR #446, opened 2026-09-16
Publish the key map's own annotation page

danvk:
The key map is georeferenced like any other sheet — by street GCPs, through `georef --geocode_keymaps` — and that pose was written to `raw/<stem>.georef.json` and never published. It is a map in its own right, and the one sheet that shows how the volume is laid out. OIM publishes these; there was no reason we could not.

`fit` now writes `<tag>.keymap.iiif.json` beside `<tag>.iiif.json`:

```
Madison, Indiana | 1904 | mapsnap generated fit (2026-09-16)              22 items
Madison, Indiana | 1904 | key map | mapsnap generated fit (2026-09-16)     1 item
```

**No stitching needed for multi-key-map volumes.** An annotation page already holds many annotations — that is how the volume's own sheets work — so Brooklyn 1939 vol 1, LA 1949 vol 14 and Kansas City 1951 vol 4 each publish both of theirs and the viewer composites them. `--label-note` keeps the two pages apart, since they are otherwise identical.

**Masking is deliberately not in this change.** `make_annotation` takes its mask in geographic space, while a key map's footprint (`regions.panels.json`) is in pixels, so it is a real piece of work rather than a flag. It can wait, because both `raw/*.georef.json` and `raw/*.regions.panels.json` are already uploaded — a mask can be added later by regenerating this file, without re-running any of the expensive chain. `--centerlines` is not passed either: the block-based clip masks are built for street sheets.

**A truth volume gets no key-map page, and says so** rather than writing an empty one:

```
No key-map annotation written: nothing in main.iiif.json provides a canvas for the key-map sheet.
```

A reference annotation page carries canvases only for the sheets it georeferenced, and the key map is not among them. A mirrored volume has no such limit — its canvases come from `metadata.json`, which lists every page — so the corpus run, which is what this is for, publishes them. Verified against a real corpus item (`sanborn02404_004`, Madison IN): one annotation, canvas `.../02404_1904-0001/info.json`, correct label.

Part of #354.


danvk:
I think Claude got confused about stitching: regions.panels.json is irrelevant, multiple keymap scan be stitched block-by-block, just like any other set of maps.

The truth volumes are different because OIM publishes their keymaps only in a separate IIIF file.

Issue #448, opened 2026-09-16
Replace the hand-rolled fleet with a Docker image and AWS Batch


danvk:
> **Revised 2026-09-18.** The first version of this issue proposed Docker + EC2 Spot Fleet and argued *against* AWS Batch on the grounds that our queue semantics were ours and would not map onto Batch's job model. That argument did not survive the test-200b run: every one of the semantics it defended is something that broke and had to be patched this week. The Docker half stands unchanged. The Spot Fleet half is superseded by Batch, which includes it. The retracted section is kept at the bottom for the record.

We have written **1,335 lines** of instance orchestration, and an instance spends **5.5 minutes booting before it does any work** (launch 19:29:07 → first item 19:34:54). Much of that is machinery AWS already sells. This issue proposes replacing it with a Docker image and AWS Batch.

Not before the corpus run is *decided* — see Sequencing, which now presents that as a real choice rather than a foregone one.

## What we built vs. what exists

| ours | off the shelf |
|---|---|
| `bootstrap.sh` (175 lines: apt, git clone, `uv sync`, EasyOCR weight prefetch, CUDA wheel swap) | a Docker image |
| `launch.sh` (229) + `shards.sh` + spot zone rotation + capacity fallback | a Batch compute environment with an instance-type list and `SPOT_CAPACITY_OPTIMIZED` |
| `supervise.sh` (158) + a crontab | Batch scales the compute environment to the queue |
| `work_queue.py` + `QueueSource`: leases, heartbeats, release-vs-retire, in-flight polling | Batch job states and dependencies |
| S3 log upload every 2 min + EXIT trap | CloudWatch Logs, one stream per job attempt |
| `_craft/done/` markers | job status |
| self-termination when the queue drains | compute environment scales to zero |

## The evidence that this is costing us

**Bugs that a baked image makes impossible.** #443 was an instance that lost three apt attempts to `unattended-upgrades` and powered off having done no work. That failure mode only exists because we build the environment at boot.

**Capacity failures that an instance-type list solves.** On 2026-09-16 we could not launch `c6i.2xlarge` spot in *any* us-west-2 AZ, and 6 of 11 spot instances were reclaimed. Spot placement scores explain it — asking for one instance type scores 1–3 out of 10; asking for a diversified 14-type fleet scores **9**:

```
                          target 256 vCPU    target 512 vCPU
single type (any)              1-3                1-3
14-type diversified fleet       9 (us-west-2, us-east-1, us-west-1)
```

[...]

PR #474, opened 2026-09-18
Dockerfile: the corpus worker as an image

danvk:
Everything `scripts/loc_craft/bootstrap.sh` builds at boot — apt, the AWS CLI, `uv sync` from the lockfile, the CUDA-or-not torch decision, the EasyOCR weight prefetch — built once. This is step one of #448 (Docker before Batch), and the step that pays for itself either way: it removes the 5.5-minute boot, the apt lock race (#443), and every "which ref did this instance clone?" question.

### Built and run

The default image is now the **CPU** one (commit 1cd8858: torch from the PyTorch CPU index on Linux; see the measurement comments below). From the arm64 Mac, `--platform linux/amd64` (emulated), then smoke-tested:

| | CPU image (default, 1cd8858) | CUDA image (5e4b87c, for `loc-craft`) |
|---|---|---|
| build context | 93 MB | 93 MB |
| image | **2.3GB** | 6.35 GB |
| torch | **torch 2.12.1+cpu**, `cuda_available=False`; no `nvidia/*`, no `triton` | 2.12.0+cu130 + 4.6 GB of CUDA stack |
| CI cold build | **182 s** | 592 s |
| entrypoint | `mapsnap` — `docker run mapsnap:1cd8858 loc-fit --queue … --run-tag …` | same |
| EasyOCR | `craft_mlt_25k.pth english_g2.pth` baked | same |
| models/ | 6 `.pt` files in, the `.bak` out | same |
| env | `MAPSNAP_GIT_SHA=1cd8858…`, `OMP_NUM_THREADS=1`, cwd `/app`, `aws-cli/2.36.48` | same |

`.dockerignore` keeps out `data/` (19 GB), `.git` (963 MB), `.venv`, `app/`. Layers are ordered so a code change rebuilds only the last one: dependencies come from `pyproject.toml` + `uv.lock` before any source is copied. A GPU image is the explicit case: `--build-arg TORCH_INDEX=https://download.pytorch.org/whl/cu126`.

### Three decisions the evidence forced

**No `.git` in the image.** It's 963 MB. The commit arrives as `--build-arg GIT_SHA`, and `git_head_info` now reads `MAPSNAP_GIT_SHA` when it isn't inside a checkout — there's no `git` binary in the image either — so provenance records the sha instead of `null` on every item the image fits. Tested.

**One torch thread by default.** A container sees the *host's* core count, so torch's own default oversubscribes a box running several jobs — the bug #444 fixed in `bootstrap.sh` by splitting cores between workers. A Batch job definition granting N vCPUs sets `OMP_NUM_THREADS=N` in its environment.

**The lockfile's CUDA 13 torch by default**, which needs driver ≥ 580 on the host. `--build-arg TORCH_INDEX=https://download.pytorch.org/whl/cu126` reproduces bootstrap's older-driver swap at build time, since the driver check can't run inside a build.

### What this does not do yet

It doesn't change how the fleet launches. `bootstrap.sh` still builds at boot; the Batch job definitions in #448 are what would consume this image. Next step after merge: push to ECR and run test-200 through Batch as the pilot #448 describes.

Refs #448

### CI

`.github/workflows/docker.yml` builds the image and runs the CLI in it two ways — `docker run mapsnap:ci --help` (the entrypoint, as a Batch job definition invokes it) and `docker run --entrypoint uv mapsnap:ci run mapsnap --help` (through uv and the venv). Separate workflow so its cost sits on its own line. Measured, four runs:

| run | trigger | cache readable | `build took` | torch |
|---|---|---|---|---|
| [35385806237](https://github.com/danvk/mapsnap/actions/runs/35385806237) | `pull_request`, first push | none | **592 s** | downloaded |
| [35386812969](https://github.com/danvk/mapsnap/actions/runs/35386812969) | `workflow_dispatch` on branch | none (PR cache is under `refs/pull/474/merge`) | **583 s** | downloaded |
| [35387896577](https://github.com/danvk/mapsnap/actions/runs/35387896577) | `workflow_dispatch` on branch, again | the 3.2 GB run 2 wrote | **342 s** | 10 layers `CACHED` |
| [35388307433](https://github.com/danvk/mapsnap/actions/runs/35388307433) | `pull_request`, **second push** | `refs/pull/474/merge` from run 1 | **224s** (success) | 10 layers `CACHED` |

The warm floor is not the download: the cached 3 GB torch layer comes back from GHA storage at ~11 MB/s (263 s), then the 6.35 GB image is exported and loaded (~130 s). The cache skips `uv sync`'s install work, not the bytes. Full breakdown and the cache-scoping finding are in [the measurement comment](https://github.com/danvk/mapsnap/pull/474#issuecomment-5735434670).

`push: branches: [main]` is on, so main seeds the cache a new PR's first run can read. Two follow-ups left to your call: gate the job with `paths:` (Dockerfile, .dockerignore, pyproject.toml, uv.lock, mapsnap/**) so PRs that don't touch the image cost nothing, and optionally make main the only cache writer. Recommendation and trade-offs in the comment.


danvk:
## Docker CI: measured

Three runs of `docker.yml`, same Dockerfile, same runner class:

| run | trigger | cache readable | `build took` | torch |
|---|---|---|---|---|
| [35385806237](https://github.com/danvk/mapsnap/actions/runs/35385806237) | `pull_request` | none | **592 s** | downloaded |
| [35386812969](https://github.com/danvk/mapsnap/actions/runs/35386812969) | `workflow_dispatch` on `docker-image` | none — the PR's cache is under `refs/pull/474/merge`, invisible to a branch run | **583 s** | downloaded |
| [35387896577](https://github.com/danvk/mapsnap/actions/runs/35387896577) | `workflow_dispatch` on `docker-image` | 3.2 GB the run above wrote | **342 s** | 10 layers `CACHED` |

Job overhead outside the build step is ~20 s; `uv run mapsnap --help` inside the image is 3–5 s.

### Where the warm 342 s goes

From the buildkit step timings of the warm run:

```
263.4s  #19  sha256:ff2c37…  3.00GB / 3.00GB     <- pulling the CACHED torch layer from GHA cache storage
 73.4s  #20  sending tarball                       <- exporting the 6.35 GB image to the daemon
 56.2s  #21  loading layer …                        <- --load
```

**The cache does not skip the torch bytes; it skips the install.** GHA cache storage delivered the 3 GB layer at ~11 MB/s — slower than the cold run's download from the PyTorch index — and the 250 s saved is `uv sync`'s resolve + install + bytecode compile, not the transfer. The floor for any run that produces this image on a runner is roughly *pull-or-download 3 GB + materialize 6.35 GB*, which no trigger filter changes.

### Cache scoping (why run 2 was cold)

GHA caches are ref-scoped; a run reads its own ref's entries and the base branch's, nothing else. Each writer costs ~3.2 GB of the repo's 10 GB (LRU). As of this PR: `refs/pull/474/merge` 3,235 MB, `refs/heads/docker-image` 3,235 MB, `refs/heads/main` 62 MB (Python CI only). The `push: branches: [main]` trigger added in the last commit is what seeds main, so a *new* PR's first run can be warm.

### Recommendation

- Keep the job, but **gate it with `paths:`** (`Dockerfile`, `.dockerignore`, `pyproject.toml`, `uv.lock`, `mapsnap/**`) on both triggers. A PR that doesn't touch the image's inputs costs nothing; one that does costs ~6 min warm / ~10 min cold. Don't make it a required check, or a non-matching PR waits forever.
- Optionally make **main the only cache writer** — `cache-to: ${{ github.event_name == 'push' && 'type=gha,mode=max' || '' }}` — so PRs never evict each other. Cost: a PR that changes `uv.lock` rebuilds that layer cold on every push.
- The real lever on the floor is **image size**: a CPU-only torch image for `loc-fit` would be ~1 GB of torch instead of 3, cutting both the cache pull and the load — and it would cut Batch's per-job image pull the same way. Worth deciding under #448 rather than here.

danvk:
## CPU image: measured on CI

Same workflow, commit 1cd8858 (torch from the PyTorch CPU index on Linux), native amd64 runner, **cold cache** — the lockfile changed, so the dependency layer could not hit:

| | CUDA image (5e4b87c) | CPU image (1cd8858) |
|---|---|---|
| image size | 6.35 GB | **2.3GB** |
| cold build | 592 s | **182s** |
| torch wheel | 2.12.0+cu130 (+ nvidia-\*, triton) | 2.12.1+cpu |
| run | [35385806237](https://github.com/danvk/mapsnap/actions/runs/35385806237) | [35390016830](https://github.com/danvk/mapsnap/actions/runs/35390016830) |

The warm number was bounded by moving a 3 GB layer and loading 6.35 GB; both shrink with the image, so a warm run of this image should land well under the 224–342 s measured for the CUDA one. Next PR push will show it.

`nvidia/*` and `triton` are gone from the lock entirely. A GPU image is now the explicit case: `--build-arg TORCH_INDEX=https://download.pytorch.org/whl/cu126`, and `bootstrap.sh` installs the same wheel on any instance with a working `nvidia-smi`.

Issue #498, opened 2026-09-21
Volume scale is estimated from one or two fitted pages with no plausibility bound


danvk:
A volume's scale is a median over its *fitted* pages, with no plausibility
bound. In a volume where only one or two pages fit, that median is whatever
those pages' RANSAC produced, and the pipeline then trusts it as a population
parameter. Three items on the 2026-09-20 corpus sample took their container
down with it; others publish a bad georeference in silence.

## The corpus prior this ignores

Volume scale across 86 published items from the sample, read from the
`volume-median` entry in each item's `artifacts/osm_snap/candidates.jsonl`:

| | m/px |
|---|---|
| median | 0.2068 |
| p10 / p90 | 0.2003 / 0.2210 |

| | share |
|---|---|
| within 0.15-0.30 | 79/86 (92%) |
| within 0.10-0.50 | 81/86 (94%) |

The distribution is tight enough that a bound is easy to place and hard to
place wrongly. Nothing currently uses it.

## Items outside it

Every outlier in the sample is a volume with one to three pages for snap to
target, which is exactly the case where a median over fitted pages is a median
over one number.

| item | volume scale | folds from 0.203 | snap targets fitted | outcome |
|---|---|---|---|---|
| sanborn09046_002 | 38.1 m/px | +7.6 | 1/2 | container killed |
| sanborn01711_002 | 14.1504 m/px | +6.1 | 1/1 | container killed |
| sanborn03404_002 | 6.39 m/px | +5.0 | 1/5 | container killed |
| sanborn05857_001 | 0.0117 m/px | -4.1 | 1/1 | **published** |
| sanborn04269_003 | 1.1049 m/px | +2.4 | 2/3 | published |
| sanborn01595_003 | 1.0533 m/px | +2.4 | 1/2 | published |

`sanborn05857_001` is the case worth worrying about. A scale that is too
*coarse* asks `osm_rasters` for an enormous frame and dies loudly, which is
what #497 now catches. A scale that is too *fine* asks for a small one, snaps
against a frame covering a few city blocks, and publishes without complaint.
Nothing flags it.

`sanborn01711_002` is the whole loop in one sheet. It is a one-sheet volume,
so `volume_median_scale` returns that sheet's own fitted scale, 14.1504 m/px.
Its page is 1613x1913, so snap asked for a frame 18,400 px on a side to go and
check the very page that set the number. The annotation it publishes places a
sheet covering well under a kilometre as one 23 km across.

## Two estimators, neither bounded

[...]

PR #499, opened 2026-09-21
Size the snap frame budget from the container, not from a guess

danvk:
#497 capped the search frame at 12,000 px on the theory that `snap` holds six
frame-sized arrays at about 15 bytes a pixel. The corpus run disagrees: 16
items have died in `fit` at exit 247 and **the guard has not fired once** in
the whole run.

## The measurement

`sanborn00138_001`, reproduced locally on `011e9d9`:

| | |
|---|---|
| volume median scale | 3.2563 m/px (16x the 0.203 corpus median) |
| frame it asked for | 6,164 px — under the old 12,000 cap, so the guard stayed quiet |
| peak stage RSS | **10,179 MB** |
| ordinary volume (~3,150 px frame) | ~3,000 MB |

That works out to 268 bytes a pixel, not 15. The six resting arrays are not
the cost; `match_at_rotation`'s correlation surfaces and rotated copies are,
and they are allocated per rotation prior and scale prior.

## What changed

The budget now comes from the cgroup memory limit the process runs under,
which is what actually kills it. The host's RAM is the wrong number — it is
many times larger and shared with every other job on the box.

| job definition | budget | frame memory |
|---|---|---|
| `mapsnap-loc-fit` (7500 MB) | 4,434 px | 5.5 GB |
| `mapsnap-loc-fit-large` (16384 MB) | 6,553 px | 12.0 GB |
| no limit (dev machine) | 12,000 px | — |

`MIN_FRAME_PX = 4_000` is a floor no container can push below. The largest
frame measured across the 30 local volumes is 3,152 px a side (Champaign 1915,
at a 2 km search radius), so legitimate work always fits: quietly declining to
snap every page would be a far worse failure than risking one kill.

A useful consequence is that `-large` earns a frame wide enough to *finish*
`sanborn00138_001` at 6,164 px rather than skip it, which is what the mop-up
pass wants from it.

## Scope

This bounds the symptom. A volume whose scale is wrong by four folds still
publishes a georeference wrong by four folds — that is #498, and it is the fix
that would stop these frames being requested at all.

Tests cover the budget arithmetic at both job definitions, the floor, and the
sentinel an uncapped cgroup reports instead of a limit.


danvk:
Superseded by the page-extent cap: capping the ground a page could cover is machine-independent and says something true about Sanborn sheets, where a pixel budget derived from the container would make the same volume behave differently on a 7500 MB job and a 16 GB one. Measured, the two are cleanly separable — largest real page 3.08 km (Hudson p92 [1]), smallest crashing scale 10.9 km.

